PR #70, opened 2015-02-25
Update creating-network-diagrams-from-historical-sources.md

mduering:
Addressed these issues:

A few minor suggestions:
Can we incorporate the “final points” into relevant sections of the text?

Rather than simply link to wikipedia for explanations of certain terms, can you offer a basic explanation as well (we can keep the link, but it shouldn’t be necessary for readers to leave the tutorial to learn basic vocabulary or concepts). just a sentence or maybe two will suffice.

I would highly suggest discarding the Roots in Qualitative Data Analysis section. Since the tutorial is directed at individual researchers, I’m not sure this adds much to the tutorial and it certainly makes it more intimidating and seem more complicated than it needs to.


fredgibbs:
Thanks for these revisions, Marten. For the final points, I meant not just to move them as a group earlier in the lesson, but mention the individual points in the most relevant section of the lesson. That strikes me as most useful, but if you're most happy with them as they are, we can leave them alone.

PR #76, opened 2015-03-19
add COinS to lesson pages

mdlincoln:
Adding [COinS metadata](http://ocoins.info/) to each lesson will, among other things, let users of the Zotero browser plugin seamlessly download all the author and contributor metadata for a given lesson.


wcaleb:
Great idea. Thanks!

PR #78, opened 2015-04-13
added <a> slugs and jumplinks

todrobbins:
for sharing specific sections of the Lessons page with others

### DIFF
--- a/lessons/index.md
+++ b/lessons/index.md
@@ -6,7 +6,19 @@ title: Lesson Directory
 We invite you to browse around. If you can’t find a skill, technology, or
 tool covered here, please let us know!
 
+- [Application Programming Interfaces (APIs)](#apis)
+- [Data Management](#data-management)
+- [Data Manipulation](#data-manipulation)
+- [Distant Reading](#distant-reading)
+- [Getting Ready to Program](#getting-ready-to-program)
+- [Mapping and GIS](#mapping-and-gis)
+- [Network Analysis](#network-analysis)
+- [Omeka Exhibit Building](#omeka-exhibit-building)
+- [Web Scraping](#web-scraping)
+- [Introduction to Python](#introduction-to-python)
 
+
+<a name="apis"></a>
 Application Programming Interfaces (APIs)
 -----------------------------------------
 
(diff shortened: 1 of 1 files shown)


wcaleb:
Thanks for submitting this! The Jekyll rendering engine we are using actually already creates slugs for every header, though. For example, this in the Markdown:

```md
## Application Programming Interfaces (APIs)
```

automatically becomes this in the HTML:

```html
<h2 id="application-programming-interfaces-(apis)">Application Programming Interfaces (APIs)</h2>
```

That makes the `<a>` slugs superfluous because you can then link to the section using that header ID, [like this](http://programminghistorian.org/lessons/#application-programming-interfaces-(apis)).

Admittedly, the slug is longer and more complicated, but it might further complicate things to have two different slug ids. It might also help if we were to make this section "link ability" feature more visible, as you've done with the table of contents. We might be able to add something that would make a permalink icon reveal itself when hovering over the section headers. That could be preferable since part of our current directory redesign was intended to highlight the graphics that @ahegel found and make the page more visual and less text heavy.

I'm open to different approaches here, though: what do the two of you think @todrobbins @miriamposner ?

wcaleb:
An example of what I mean about the hover icons:

http://ben.balter.com/2014/03/13/pages-anchor-links/

todrobbins:
Great points! I thought Jekyll did that but I couldn't figure out the pattern. Thanks for highlighting that.

The automatic slugs are fine, and I think Balter's hover icons CSS/JS would be a lot better than what I threw together. :smile:

miriamposner:
Sounds good to me! Sorry, preemptively merged there. Thanks, Caleb!

On Mon, Apr 13, 2015 at 2:37 PM, W. Caleb McDaniel <notifications@github.com

wcaleb:
Thanks, no worries! It calls attention to a usability issue we definitely should address.

todrobbins:
Do you want me to create a new issue about integrating the hover icons? Or is that something the core team would rather spearhead/manage?

Regardless :+1:

wcaleb:
Sure, it wouldn't hurt to create an issue so that it eventually gets done!

PR #85, opened 2015-04-27
Took out "Lessons We Want"

miriamposner:
It seems out of date and incomplete. An alternative might be to start a "wish list" on the Wiki and just link directly to it. What do you think?


wcaleb:
Seems like a good idea to me!

PR #88, opened 2015-05-02
Suggested citations

### LINKED ISSUE #82: Suggested citations
I think we need to add suggested citations for our lessons and for the project. This should be possible via the YAML files, no?

mdlincoln:
Closes #82


wcaleb:
Thanks!

wcaleb:
FYI, @mdlincoln, in d4260595f709c4ebd877920a0d24c860a072c35a I abstracted the `authorString` variable used both in the byline and in the suggested citation so that it adjusts for one, two, or more authors and puts commas and "and" in the right place.

PR #90, opened 2015-05-09
future proofing for changing between inline and footnote styles

denten:
A few minor changes to give better examples for citations inside and outside quotes. I tested this with pandoc 13 / citeproc 7 and it seems to work well in transitioning between footnote and inline styles.


ianmilligan1:
These minor changes are great and help with the lesson flow. Thanks for these updates, Dennis, much appreciated!

PR #106, opened 2015-06-12
New lesson: "Installing Omeka" 

JonathanReeve:
This is a reworking of a lesson I wrote last year, which deals with manually installing Omeka on a web server. It presumes no prior knowledge of the command line, but uses command line utilities to install LAMP, create a database, and configure all the database settings in order to get a working Omeka install. Let me know what you think!

I also added an author bio, and uploaded images.


fredgibbs:
Hi Jonathan,

Thanks for submitting this new version of your lesson! I've pulled it into the repository. Before we send it out for review, I want to get your thoughts about one potential change. It seems like the lesson is constantly differentiating between two kinds of hosting setups (VPS and shared). It seems like it might be clearer to present two separate sets of instructions, so new readers can simply identify what kind of hosting setup they have and proceed accordingly. This would take a bit of reorganizing, but shouldn't involve much rewriting. I think it would make your nicely articulated steps much easier to follow. What do you think?

Thanks again!
fred

JonathanReeve:
Hi Fred,

Good idea. I'll see if I can reorganize it into two sections, one for VPSs, and one for shared hosting. I've just opened [an issue for this](https://github.com/JonathanReeve/programminghistorian/issues/1) and will get to it shortly. 

-Jonathan

JonathanReeve:
@fredgibbs: I just opened PR #230 with some revisions following your suggestions. Let me know what you think.

PR #136, opened 2015-08-27
Fix minor typos.

adunning:


ianmilligan1:
Thanks, @adunning! Good catch on these minor typos.

adunning:
Thanks for the fantastic resource!

PR #139, opened 2015-09-14
Create Introduction-to-R.md

taryndewar:


wcaleb:
Live preview of lesson is here: https://proghist-dev-pr-139.herokuapp.com/lessons/Introduction-to-R.html

williamjturkel:
@wcaleb Should I open a "Review of Introduction to R" issue now, or should I accept the pull request first?

wcaleb:
Don't accept pull request. Instruct reviewers to leave comments directly on this pull request comment box. Then when the author updates the lesson on their branch, it will be reflected in the pull request and in the live preview. Accepting the PR is the now final step. 

Sent from my iPhone

W. Caleb McDaniel
Associate Professor of History
Duncan College Master
Rice University

http://wcm1.web.rice.edu/
http://www.duncancollege.net/

histlib:
This is John, doing a first chunk of comments. I'll refer to paragraphs by section, paragraph number, & sentence (e.g., Intro 2,1 = first sentence of the second paragraph of introduction section).

Intro 1,1: "quantitative" rather than "qualitative"
Intro 1,3: This sentence is likely to be over the heads of intro users. I get that this sentence is part of the simple/complex binary from the opening, but it might be more helpful to reshape the paragraph as a whole around showing the reader why they should be excited about R and the range of ways it gets used (especially in history or the humanities).
Intro 2,2: "analyze and visualize the data" rather than "generate information about"
Intro 3,2: in the third clause (starting with "as well as"), there needs to be a verb to parallel "take you through"/"explain" formula (e.g., "and show you how to work with" ...)

Installing 1,2: why link to the Berkeley mirror rather than https://cran.r-project.org/?
Installing 1,3: re: compatibility with OS - at this point, insert comment about which OS(s) this tutorial will address (since R in Linux is deployed directly in terminal, guessing that's not the focus here)
Installing 2: image link doesn't work
Installing 3,1: since some (or all) of these terms will be unfamiliar to a novice (though they are explained later), is there a simpler way to describe what the console does? It might just be a matter of "The console is where you will type in your commands."
Installing 3,3: my Windows console doesn't have a colour wheel - I have to go to Edit -> GUI preferences
Installing 4,1: help() opens up the help page for the function help; the help function is usually used in conjunction with a package name (as a character string); general help can be obtained using the Help menu at the top of the console (and has the benefit of having help for using the console, plus links out to CRAN, FAQs, manuals, etc). help.start() takes one to an HTML page to access the manuals
Installing 4: it might be useful to have a separate section on how to look for help, especially since often the best way to trouble-shoot an error is to do an Internet search for the error message (R manuals are often terse and unhelpful to a new learner)

drjwbaker:
Specific comments:
- 'linear and non-linear modelling'. You should explain what this is (or just leave it out as the lesson does not cover it).
- 'This tutorial will go through some of the basic functions of R and serves as an introduction to the language.' But why would historians want to read this? You need to make the sell more clear
- 'R runs like a command-line system where you enter data into a console row by row'. Probably worth noting here that it can just be run from a command-line if preferred (which links into existing lessons such as - with apologies... - Milligan & Baker, '[Introduction to the Bash Command Line](http://programminghistorian.org/lessons/intro-to-bash)'
- 'There are two ways to go about doing this – you can either select the terms you want or create a variable to represent those terms:' Which have you demonstrated here? It is not clear to me
- `y1955<- AirPassengers[73:84]` seems a clumsy way of defining a variable as you need to count through the data to find the first and last entries for 1955, that is 73 and 84. Thus it doesn't scale well. Could not a function such as `filter` be used to select the row required?
- `> summary(AirPassengers[73:84]` missing a `)`
- 'If you only needed to use the data once, it would probably be easier to use the second option, but if you were doing a number of different calculations, creating a variable for each year would make more sense.' Why? I'd tease out the decision making here as it is an important part of using a language

General comments:
- I am concerned that this lesson is going about things the wrong way round. Historians will have data and if they find this lesson will want to know how R can help them. As this lesson rightly suggests, they should be guided to use sample data (data that works!) before throwing their own at R. However I suspect most won't care about the maths functions available in R in an abstract sense. So I'd start (after the pre-amble) with the datasets bit and then use the sample datasets to explain the basic maths functions and matrices (the bits currently at the beginning) through worked examples with the sample data.
- Given that you mention creating graphics in the intro you should either a) state that you won't be covering graphs (and point folks at guides in the further reading, such as [Lincoln Mullen's work](http://lincolnmullen.com/projects/dh-r/)] or b) add some very simple viz as a taster
- For me, an important reason for 'converting' historians to R is that it enables them to reproduce what they do. That is, they can re-run the same code and get the same results, the same viz. I'd stress the robustness of R and the role it can play in making research data analysis reproducible. I'd also point the reader to those bits of the R Console that enable historians to go back to old functions to either repeat them or build on them (I use R on the command line, so not sure where you'd find these in the Console if at all!)

taryndewar:
Thank you both for your comments! I appreciate the time you have given to improve the lesson. This will definitely help to strengthen the lesson and make it more accessible to the target audience.

acrymble:
Can we clarify who the editor is for this lesson?

acrymble:
As Bill Turkel as stepped down as an editor at the Programming Historian, I will be taking over as editor on this lesson. As this lesson has already undergone a review, my role is to mediate between reviewers and authors, and keep the process on track in a timely manner, etc. If anybody needs to contact me outside this forum, my e-mail is adam.crymble@gmail.com, although in keeping with public scholarship and peer review we generally encourage discussions to take place on this pull request.

If anyone feels that they have a matter to discuss and are not comfortable raising it with me, or need a second opinion, you can contact either of our ombudspersons, Miriam Posner or Ian Milligan (http://programminghistorian.org/project-team).

I will need a few days to look through everything here, including the lesson, but I'll be in touch as soon as I can.

acrymble:
Ok, just to clarify, Taryn is working away on these revisions, so I'll leave her to it!

acrymble:
Taryn is revising and hopes to have an update in a few weeks.

acrymble:
this has been moved to our submissions re po.

PR #165, opened 2015-12-07
Create transforming-xml-with-xslt.md

mhbeals:


acrymble:
Thanks @mhbeals I will take on the role of editor for this. In the first instance, I'll have a read through and try the lesson and initiate a dialogue with you about the initial submission. After you've had a chance to respond, I'll find a couple of reviewers to try it as well and I will facilitate that conversation to ensure that it remains productive and timely.

I will endeavour to keep the conversation open here on Github. If you feel the need to discuss anything privately, you are welcome to email me. You can always turn to @ianmilligan1 or @miriamposner if you feel there's a need for an ombudsperson to step in.

Looking forward to working with you.

mhbeals:
Thanks @acrymble. I have also synced the images with the images folder, but seem to have done so to the main branch (for some unknown reason) and therefore they have not been included in this pull request.  They are available at: https://github.com/mhbeals/jekyll/tree/gh-pages/images.

acrymble:
Dear Melodee,

Thanks for this lesson. I think there's a lot of potential here for this to become the go-to place for historians and history students to learn the power of XSL. This lesson would be particularly useful for a classroom workshop, which is fantastic.

There are a couple of areas that I think need some additional work. Broadly these fall into two categories:

1) Expanding the potential audience by lowering the barrier to entry
2) Fixing/Adjusting a number of small points

---

1) Barrier to entry:

With a few exceptions outlined below, I managed to get through this lesson without any big problems. But I think we want to reach a wider potential audience than people like me. For example, people who don't have as much experience, or who aren't familiar with the terminology or the technology.

There's quite a bit of assumed knowledge in the tutorial, including a lot of jargon - though I really like your explanation of terms like 'comment-out'.

I think the introductory bits could use some significant flushing out. Namely: think of your skeptical, technologically inept colleagues who have no idea what XSL is or why they should care. After reading a couple of paragraphs they should have a clear idea of what problem they face as historians that will be solved by this technique.

At the moment, I think we don't quite get that, in part because it's implied that the reader will know a fair bit about XML. We know the value of XML, but most people have no idea what it is, why anyone would use it, or even the basics. An XML example file (in code blocks) right at the top would be helpful. Also noting that often you’ll come across XML other people have made, not just your own. We want people to know this is a broadly useful skill, and isn’t just for XML junkies. Tell us what an XML file is (nested tags that isolate bits of information within a text, etc).

I think it would be very helpful to start with a historical problem rather than a technical explanation as you have now. Eg, you have 1000 XML files containing a,b,c, and you need to get all of the ones with fuzzy peaches in them. How do you do it? Use XSL!!! Something quite general that you then solve in the lesson. I think it would help to significantly lower the barrier to entry, and keep people reading if they're not sure this is what they need.

2) Fixing / Adjusting a number of small points

There are a number of small issues / ideas for tweaks I came across, some of which are stylistic, others are jargon that just need a bit more context. I've included those here as a list:
- LN 2: The title isn't going to attract a historian. Any ideas to make this a bit more historian-focused w/o losing the important bits?
- LN 2 and various: XSL or XSLT? Be consistent. Keep in mind that XML and XSL look very similar.
- LN 14: Is your XML database really a database, or is it a flat file? The answer might be yes, but this just struck me as odd. I suppose it's a 'document-oriented database'.
- LN 14: What do you mean by 'a variety of different formats'? How would a historian understand that?
- LN 17: XSL doesn’t provide XML files with instructions. It provides instructions for another program to interpret and rearrange the XML for display.
- LN 32-36: The way you word your last line, I wasn’t sure what I needed to download. Firefox? Or a plugin? The next line doesn’t look like instructions for helping me decide what I’d be most comfortable with. Can you add a simple segue line at the end of this section that says something basic like ‘once you have installed a text editor and chosen a browser, move onto the next section’…?
- LN 39: What is the Scissor and Paste Database? We want to know what we’re working with so we care about why we’re doing this transformation to it. Eg, 23000 lines of newspapers… lots of stuff we don’t necessarily want to look at… blah blah (plug your research).
- LN 47-52: I was confused here because you didn't mention all of the other files. It makes the reader wonder if they've downloaded the wrong thing. I would mention the read me, license, and cite files. It’s a bit confusing to read there are 3 things and then see 7. But I also think this is a great subliminal message to readers that they should distribute their own data with these types of things.
- LN 58-81: This should be your opening bit. Here is our historical source, but it's not how we want it. There's your historical problem, and your chance to explain XML to the readers.
- LN 66-79: Tell reader that you came up with the elements from amongst the options available from TEI. Otherwise they might worry that their XML is wrong or won’t work with this lesson. Make it clear that YOUR XML isn’t crucially tied to the outcomes and that this will work on THEIR data too.
- LN 83, 91, etc: Note that the name of the XSL file isn’t important, but that’s the one you chose for the example.
- LN 91: “(that is, your desktop.)” – actually it’s in ‘Simplified’, which is on my desktop. Someone might get caught on that.
- LN 121: Figure 7 has the following line at the end: [code]/xsl:stylesheet[/code] I don’t think you told us to add that. This was causing an error for me. Using code blocks to show what the reader should have at various points would avoid this problem.
- LN 134: F5 doesn't refresh on my mac. It brightens the screen. Refresh, or cmd r work. This is machine dependent instruction.
- LN 162: The line break character doesn't work on my mac. It just does nothing at all. Not sure why.
- General: More use of code blocks rather than screen shots to show the reader what their code should look like. This lets people cut and paste if they're having trouble. You do this at the end, but not when you're building up your initial example.
- Various: there are a lot of jargony words. Generally I encourage links to wikipedia for these. The ones I came across that would benefit from this include: XSL, XML, HTML, CSS, WYSIWYG, text editor, word processor, non-ascii, top-level element, namespace, match attribute, root element, syntax errors, ISO date. Some of these may need a bit more explanation for a novice reader.
- Bit I got stuck on: I accidentally tried dragging the XSL instead of the XML into the browser. And it didn't work. I misread, but probably others will too.

---

Please have a read through these comments and ideas and then let me know what you think. This is an opportunity for a conversation, and I hope some of these ideas will help build in the strengths of the lesson and make it appropriate for an even wider group of potential readers. I think this one could be a hit, particularly in the classroom. So it's best to get the level right.

mhbeals:
Many thanks for these helpful suggestions. The finer adjustments are all easily implemented (and some extremely well spotted!) changes, which I agree will make the example tasks more straightforward across platforms.

In terms of the first section, I am happy to add a more detailed narrative, which may provide better context and a sense of wider applicability to PH browsers. However, I am unsure to what level of detail and format would be ideal. Those undertaking XSL will need a basic understanding of XML, but I am not sure how much of that should be integrated into THIS tutorial, as opposed to a stand-alone tutorial to creating XML data.

Which of the following, therefore, do you feel would be best in the context of the PH ecosystem
- A separate tutorial on XML construction that leads into this one
- An integrated 'basics' to XML, focusing on _using_ key resources that have XML data, rather than historians creating their own

-MHB

acrymble:
Thanks Melodee, I don't think you need to do a full lesson on XML. But it is probably helpful to give some basics to let people know that you're talking about documents that have already been structured for use with XSL. That structure has been imposed using XML, which highlights various types of information within a text or series of texts. In your case, your XML lets us programmatically find all of the titles, dates, etc. This tutorial will teach you how to extract those targeted bits of information and display only the ones you're interested in, in the way you define.

What we don't want is someone showing up with an MS Word document and wondering why they can't extract all of the references to onions, or something like that.

We can leave this for the review round if you want a second and third opinion.

fredgibbs:
+1 for your second option, @mhbeals, on presenting just the minimum basics of XML within your own lesson. It will be nice to have a separate XML tutorial that goes into some depth (and maybe creating XML files), but we'll leave that for another time. I think your focus on using data is exactly right. But as @acrymble said, the reviewers will likely have good suggestions on this issue as well.

mhbeals:
Dear @acrymble,

Thank you for your thoughtful revisions to the original text.  I have now updated the tutorial to reflect your suggestions, including a more detailed (and hopefully gentler) introduction.

Please let me know your thoughts and the next steps.

Best wishes
M. H. Beals

acrymble:
Thanks, I'll have a read through as soon as I can.

acrymble:
This is now on the submissions repo. Should I close this @wcaleb ?

PR #174, opened 2016-01-04
Introduction to Augmented Reality with Unity

jacobwgreene:
Have been working with Fred Gibbs. Images have yet to be uploaded;
awaiting confirmation.

### DIFF
--- a/lessons/Introduction-to-AR-Development-with-Unity (1).md
+++ b/lessons/Introduction-to-AR-Development-with-Unity (1).md
@@ -1,0 +1,303 @@
+---
+
+title: Introduction to Mobile Augmented Reality Development in Unity
+
+author:
+
+- Jacob W. Greene
+
+date: 2015-1-4
+layout: default
+
+---
+
+## Lesson Goals
+
+In this introductory tutorial, you will learn how to:
+
+* setup the Unity game engine for AR development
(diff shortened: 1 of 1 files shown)


fredgibbs:
@wcaleb, can we get a preview of this from the pull request? If so, what's the best way to handle images? Do the images need to be in the same pull request as the lesson (they aren't now, obviously) for the preview to work with them? Thanks for any advice!

wcaleb:
@fredgibbs There's an issue with the Heroku system that I'm not yet sure how to resolve: we get a limited number of Preview Apps with a free account, and we're using them all right now. I've started to sour a bit on using that solution for previewing anyway for other reasons, and hoping to talk about alternatives on next editor call. If you need to move more quickly on this one, I suppose we could depart from usual practice and go ahead and merge like we did in the past, but if it can wait a week, we can talk on Skype.

acrymble:
I only see 7 lessons with pull requests at the moment, and all four with me
don't have previews. Do we have old previews we can get rid of?

wcaleb:
I guess it counts disabled apps towards your quota. 

heroku

acrymble:
this is done now ya?

ianmilligan1:
I'm closing this as think this is done. Feel free to hit me if I am wrong. 😄

PR #215, opened 2016-02-26
quick fix to switch to My Maps

jburnford:
I've edited the text to direct users to My Maps instead of Map Engine Lite. It all still works. I still need to switch the images.


ianmilligan1:
Jim is sending me the images, will drop in when I'm on my system (probably tomorrow).

PR #235, opened 2016-03-07
Keep only Gemfile, not Gemfile.lock

mdlincoln:
Ruby's bundler gem generates Gemfile.lock based on Gemfile, so the former should not be tracked by git. (sorry for so many issues on this! I think this should resolve everything)


wcaleb:
Thanks!

PR #254, opened 2016-05-27
Removed dated ombudsperson reference

amandavisconti:
Removed reference to Miriam Posner as a current ombudsperson.


fredgibbs:
thanks for the catch!

PR #260, opened 2016-06-13
Added difficulty metadata (1)

amandavisconti:
Added the difficulty data to the front matter. I set this lesson as a 1 because it doesn't assume prior knowledge.


ianmilligan1:
Sounds great, @amandavisconti – makes sense to me.

PR #263, opened 2016-07-08
Updated my bio/interests/team:true

amandavisconti:
Hm, updating this page routed me to a pull request. Is that the norm (so the rest of the team knows when a change is being made to the live site), or should I have permission to commit directly?


acrymble:
Maybe caleb hasn't given you permission?

ianmilligan1:
I just sent you an invite to join the team, @amandavisconti – should be in your inbox! 😄

Issue #289, opened 2016-09-09
Broken links in Sparql lesson


bmschmidt:
There are at least two broken links in the graph-databases-and-SPARQL.md lesson. One is to http://rdf.myexperiment.org/howtosparql? at the end. More critically, the page linked under 'Europeana's' in the following quote leads to a blank page, making it hard to follow along.

---

Up until now, we have constructed queries that look for patterns in one dataset
alone. In the ideal world envisioned by Linked Open Data advocates, multiple
databases can be interlinked to allow very complex queries dependent on
knowledge present in different locations. However, this is easier said than
done, and many endpoints (the BM's included) do not yet reference outside
authorities.

One endpoint that does, however, is
[Europeana's](http://europeana.ontotext.com/sparql). They have created links
between the objects in their database and records about individuals in
[DBPedia](http://wiki.dbpedia.org/) and [VIAF](https://viaf.org/), places in
[GeoNames](http://sws.geonames.org/), and concepts in the Getty Art &
Architecture thesaurus. SPARQL allows you to insert `SERVICE` statements that
instruct the database to "phone a friend" and run a portion of the query on
an outside dataset, using the results to complete the query on the local
dataset. While this lesson will go into the data models in Europeana and DBpedia in depth, the following query illustrates how a `SELECT` statement works. You may run it yourself by copying and pasting the query text into the [Europeana endpoint](http://europeana.ontotext.com/sparql).

mdlincoln:
Thanks for submitting this.
1. I would probably just remove http://rdf.myexperiment.org/howtosparql
2. This is _sigh_ a more serious issue. It looks like Europeana is no longer supporting the SPARQL endpoint, at least not in the form it was in when I wrote the lesson:
   ![screen shot 2016-09-11 at 7 33 46 pm](https://cloud.githubusercontent.com/assets/4392604/18423438/e734cf64-7859-11e6-80c9-0307d0261d0e.png)
   I'm reaching out to Europeana Labs to check on the story behind the endpoint, but it _was_ a "beta" service, and so may truly be down for good. If this is so, I'd suggest adding some type of notice at this point in the lesson that the service has been deprecated. It might even be a good space to point out one of the common weaknesses of "linked open data" services...

Any thoughts, editors?

acrymble:
Thanks for flagging @bmschmidt. Let us know how Europeana responds @mdlincoln. That's the best first approach.

mdlincoln:
[...]

PR #290, opened 2016-09-12
Sparql link 404

mdlincoln:
Addresses #289 
1. Replaces a broken link to a SPARQL tutorial in the "Further Reading" section with a link to the W3C overview of SPARQL (which is actually a better place for more in-depth study, anyway!)
2. Adds a note to the "Linking multiple SPARQL endpoints" section explaining that the Europeana endpoint is down as of 2016-09-12 until further notice (pending new info from the folks at Europeana Labs)

@wcaleb I bolded the advisory note, but if there's some other class baked into the PH CSS that would be better to use for this deprecation note, please let me know / edit the PR

### DIFF
--- a/lessons/graph-databases-and-SPARQL.md
+++ b/lessons/graph-databases-and-SPARQL.md
@@ -462,7 +462,7 @@ ORDER BY DESC(?n)
 
 ## Linking multiple SPARQL endpoints
 
-**2016-09-12: Unfortunately, Europeana has suspended their SPARQL endpoint until further notice, so the links in the following section are no longer operative.**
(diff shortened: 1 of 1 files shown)


wcaleb:
There is a `deprecated: true` flag that [produces this message at the top](https://github.com/programminghistorian/jekyll/blob/gh-pages/_layouts/default.html#L35), but I don't know if it would be appropriate in this case. I think on some other recent lessons @acrymble has written a more descriptive message like the one that you have in this PR. Thoughts?

mdlincoln:
Agreed, I would not use the `deprecated` flag because it is inappropriate. If you're fine with just a boldface message at the start of the affected section, then we're all set. I suppose you could make it red or something, too?

PR #298, opened 2016-10-06
Had to add just one space to fix another broken heading

jjon:
Sorry all, there was still just one broken heading so this pull is to add just a single space character; geez! Is it the markdown parser in jekyll that's so picky?

### DIFF
--- a/lessons/generating-an-ordered-data-set-from-an-OCR-text-file.md
+++ b/lessons/generating-an-ordered-data-set-from-an-OCR-text-file.md
@@ -612,7 +612,7 @@ As before, run this script repeatedly, correcting your input file manually as yo
 Our text file still has lots of OCR errors in it, but we have now gone through it and found and corrected all the specific metadata bits that we want in our ordered data set. Now we can use our corrected text file to build a Python dictionary.
 
 
-#<a name="dictionary"></a> Creating the Dictionary
+# <a name="dictionary"></a> Creating the Dictionary
 
 Now that we've cleaned up enough of the OCR that we can successfully differentiate the component parts of the page from each other, we can now sort the various bits of the meta-data, and the charter text itself, into their own separate fields of a Python dictionary.
 


ianmilligan1:
Hah, no need to apologize – we're sorry for missing it in the first place. Must be?

wcaleb:
I think that space is fairly common as a requirement in Markdown parsers. You also don't need to add your own `<a name>` tags to the headings. GitHub Pages will automatically turn the header titles into anchors on the page.

jjon:
Thanks Caleb, Ian. I didn't add the anchor tags. That was done sometime during the editing process.

These heading errors are hard for me to test: github's markdown processor is permissive and all these heading errors were correctly rendered in the repo. It must be jekyll's processor that's so strict, and I don't have access to the jekyll instance that renders the website.

I had a look at the Jekyll repo's [note on the md processor](https://help.github.com/articles/updating-your-markdown-processor-to-kramdown/) where it says: 

> If you are not already using kramdown, Jekyll 3.0.0's default Markdown processor, then you must update your Markdown processor in your _config.yml file.
> 
> GitHub Pages only supports kramdown as a Markdown processor.
> 
> GitHub-flavored Markdown is supported by kramdown by default, so you can use Markdown with GitHub Pages the same way you use Markdown on GitHub.

I don't know if this may be the source of the problem or not. It's a wee problem to be sure, but I know you'll want your md processor to be consistent with the one in use on github.

wcaleb:
Thanks, Jon. We do use Kramdown. I think you're right that GitHub's rendering is, nonetheless, still more permissive, regardless of what they say! The proliferation of parsers is definitely annoying. Have a look at CommonMark if you really want to go down the rabbit hole ...

jjon:
Thanks Caleb, but one rabbit hole at a time ...

Just FYI: Memory betrays me. On thinking about it, I believe I did build the TOC by inserting the anchor tags. This was, I believe, to override Markdown's mechanism for creating TOC which insists that headings in the TOC match those in the document. Some of the headings in my document were too long for the TOC, and that's how I wanted them. So I reverted to html. I figured this out by making experiments in my repo, screwing it up repeatedly, struggling to fix it, and finally getting back to where we were (I hope).

None of that has anything to do with jekyll's finicky whitespace parsing. But since the headings are now rendered correctly, and the TOC works fine, I'm leaving well enough alone.

Thanks again Ian, for attending to this so promptly.
J

PR #349, opened 2017-02-27
Fixing Broken Link

bmschmidt:
The HathiTrust lesson appears to contain a broken file link. This one I change it to seems to be correct, although there are actually two files in the `listings` directory when you rsync it, and it's possible the authors might prefer the older one (full-listing.txt, from November).

Apologies if this is a double submit: I thought I sent this a couple weeks ago, but now think I must have just filed a pull request on my own fork, not the main site.


ianmilligan1:
Great catch, thanks Ben! Tested and verified. I'll merge, but ping @organisciak too.

Issue #370, opened 2017-03-05
"Editor" in yaml instead of under 'Reviewer'


acrymble:
As per #340 - @wcaleb has suggested converting all editors listed as reviewers to 'editor' instead. This ticket is to identify the offending lessons and start revising.

acrymble:
The following files need the editor pulled out from amongst the reviewers:

- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/applied-archival-downloading-with-wget.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/automated-downloading-with-wget.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/building-static-sites-with-jekyll-github-pages.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/cleaning-data-with-openrefine.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/cleaning-ocrd-text-with-regular-expressions.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/code-reuse-and-modularity.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/corpus-analysis-with-antconc.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/counting-frequencies-from-zotero-items.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/counting-frequencies.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/creating-an-omeka-exhibit.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/creating-and-viewing-html-files-with-python.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/creating-network-diagrams-from-historical-sources.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/creating-new-items-in-zotero.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/data-mining-the-internet-archive.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/downloading-multiple-records-using-query-strings.md
- [x] https://github.com/programminghistorian/jekyll/blob/gh-pages/lessons/extracting-keywords.md
[...]

PR #382, opened 2017-03-10
List all reviewers/editors/translators

mdlincoln:
Uses some liquid workarounds to address #370


wcaleb:
Cool! I didn't know about array_to_sentence_string. Thanks! Merging.

Issue #414, opened 2017-04-27
evaluation of ongoing design work


fredgibbs:
I'm creating a ticket for a dedicated thread to identify and discuss issues with the new css/layout/design work i've been doing at http://fredgibbs.net/jekyll/.

In particular, there is one code issue that needs to be evaluated. I have coded the nav bar to get language info from https://github.com/fredgibbs/jekyll/blob/gh-pages/_data/snippets.yml. i think this is the cleanest way to maintain translations that appear in include files, but others, esp @wcaleb and @mdlincoln should double check that this approach makes sense. If it's OK, then we will probably need the topics.yml file to be organized (and translated) in a similar way so the spanish lessons page will work like the new english one from @amandavisconti.

I have edited many of our instruction pages to streamline the prose--cumulative editing over many years has resulted quite verbose instructions! I haven't changed (at least on purpose) any workflows or policies. The biggest change was that I removed a bunch of the generic markdown instructions from the author-guidelines page since we have a tutorial on that now. Our site specific stuff is still there of course. 

Looking forward to hearing your suggestions!

mdlincoln:
Thanks @fredgibbs - I'll take a look this weekend.

acrymble:
Really love your new work here @fredgibbs .

I havn't had a chance to go through it all yet. But a few quick comments:

- [x] will your changes to the text be easy for the Spanish team to incorporate without re-translating everything?
- [x] On front page (http://fredgibbs.net/jekyll/) I find the blue and teal band too similar in colour to link colours.
- [x] re: footer. The ISSN number has been removed. This seems like the least obtrusive place to put it (though I also think that should be in the new 'about' page somewhere). I'm not sure it's relevant that we're hosted in Github. And 'See Change History' seems a bit jargony. Will enough people understand that?
- [x] On about page, the 'Editorial Board of the Programming Historian' has been removed. I appreciate you might not like that term, but there are times when we need to fill in forms that require a name of a publisher (eg, when submitting material to the British Library). I think it's a good idea to encourage consistency in that.
- [x] wondering how we can make the Spanish link more obvious without being obtrusive.

Overall, very great stuff in my opinion.

amandavisconti:
This looks amazing! I've been out moving between states and starting a new job this month, but I'll wrap up the dangling pieces of my lesson page work before the May call and let Fred know when they're ready for incorporation.

jerielizabeth:
[...]

Issue #398, opened 2017-03-27
Should published lessons be moved from 'submissions' site


acrymble:
Issue was raised over email to editorial board. Will the github submissions repository get too cluttered if we don't move completed lessons? 

How do we ensure moved lessons still work (internal links, link to github review tickets).

How can we ensure any extra work doesn't outweigh any extra benefit to moving?

wcaleb:
I suggest we put them in a "published" folder and edit the YAML to include a `redirect_from` the original URL (without `published` in the path) so that the live previews mentioned in review tickets still work.

mdlincoln:
I second moving them to a dedicated folder. This would also let them be excluded from html-proofer checks (which should eventually be added to ph-submissions like they have been to the live site).

ianmilligan1:
Thirded (?).

wcaleb:
I've run into a bit of a problem here. If I move a published lesson into a new folder on the `ph-submissions` repo, then we lose the whole commit history for the file. (Unfortunately, this means we've already lost the commit history for [the lesson that I have test moved](https://github.com/programminghistorian/ph-submissions/tree/gh-pages/lessons/published).

This raises two questions, probably for @mdlincoln:

1. Is it somehow possible for me to restore the repo such that the commit history for that file can be restored? (Quick glance at Stackoverflow suggests [the answer may be no](https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history).)
2. Is it important for us to keep the commit histories? My recollection is that we thought it was, as a way of documenting the whole revision and review process. Otherwise, there's no real reason to keep the published lessons in the ph-submissions repo at all, unless people want to come back and study the changes that were made in response to reviewer suggestions.

TL;DR - If commit history is important to us, then our plan to move published lessons into their own subfolder may have a fatal flaw.

mdlincoln:
You're right that git has rather lossy behavior for tracking moved/renamed files. The file history will be less visible on GitHub, because it uses the basic `git log` function. Someone running `git log --follow` on the other hand, can still access the full change history. I did that just now, and can see the full change history even after you moved the file. So, it's still accessible, but you need to use the command line. **That is definitely a high barrier!** But then again, for the (low?) numbers of people going through old revisions, that may be an acceptable barrier.

[...]

PR #425, opened 2017-05-08
Add instructions on Travis link checking to editorial workflow

mdlincoln:
Now that we have Travis CI running constant checks of internal and external links, it will begin catching typos on newly published lessons as soon as editors commit them to the main repository. I think it would be a help for editors to check for build errors as part of their final publishing workflow, since it has happened that URL typos in new lessons and in translations have led to some broken links making their way online.

I hope this doesn't read as an onerous addition. I added some screenshots, since navigating the Travis CI build results can be intimidating your first time around. In most cases, the tests ought to pass fine, and editors will not even need to pull up the Travis site.


mdlincoln:
FYI @jerielizabeth I just tagged you since I knew you have been working on editor workflow management - but I'm glad for comments from anyone!

acrymble:
this should be done on the submissions repo, no? Before the files get moved?

Will this also catch missing elements from the YAML?

The editorial checklist is getting rather long. I wonder if we don't need someone to act as a typesetter as distinct from editor? I imagine it would iron out some inconsistencies and reduce the burden on editors.

mdlincoln:
Good questions.

1. We could add Travis CI checking to the submissions repo as well. However, #398 needs to be finished (cc @wcaleb), so that old submissions are not constantly re-checked.
2. It will only catch formal errors in the YAML (aka someone put in an illegal quote mark), not semantic errors like leaving out a field. We could probably add an additional routine to the Travis script to check for required YAML fields, but I'd want to scope out how much work that would involve.

wcaleb:
I need to get on #398, thanks for the reminder. I would be willing to serve in the "typesetter" kind of role if we end up going in that direction. At this point scrubbing Markdown, fixing errors and the like is one of the main things I do already.

mdlincoln:
Ditto on scrubbing/tidying the text and code in this repo - I'm already doing a lot of that as well and am happy to continue.

I do think it's important for editors to be aware of these tests at the time of publishing, because it's often easier and more appropriate for editors to fix bad links at that stage in the process. They will be familiar with the lesson text and the intent/context of a link, and have closer contact with the author, and thus be in a better position to repair mistakes intelligently.

Thus, I think a "typesetter" should, in this case, be reactive, helping to catch any mistakes the editor didn't/can't fix, rather than having to assume responsibility for _all lesson metadata and formatting_. I feel like the latter would just devolve into me pinging an editor with a dozen queries for things that the editor could have just filled out in the first place - duplicating effort, rather than reducing it.

editor-guidelines.md
```diff
@@ -198,6 +198,47 @@ Upload the original image to the [gallery/originals](https://github.com/programm
...
+
+We run these builds primarily to check that URLs that _once_ were functional are _still_ functional, as often times external web pages are moved to new addresses, or are no longer live.
+They are also an excellent way to catch small typos that may have evaded authors, editors, and reviewers.
+The status of these tests (often called a "Build Status" on Travis CI and on GitHub) can be seen by navigating to the [programminghistorian repository page][ph_repo], and clicking "Commits" on the upper left side of the code menu.
+
+![GitHub commit menu location](images/editor-guidelines/gh_commits_location_screen.png)
+
+This will show you the list of every change made to the main repository, along with a status icon:
+
+- Green check mark: you're good to go! All the links on the page were checked and found valid. **You may skip the rest of this section.**
```
jerielizabeth: Quick thought: what about an internal link to the end of the section inside the "You may skip the rest of this section"?

jerielizabeth (commented):
I think it looks great! Just the one suggestion for ease of navigation for editors with error-free commits.

jerielizabeth:
If it's not too much work, I think having Travis check the required YAML fields would be useful. There are quite a few details to keep track of and I think the automated check would be the most reliable way to verify that we are all using the same fields (as opposed to another human reader.)

editor-guidelines.md
mdlincoln: good call! Will add.

mdlincoln:
Alright, I'll look in to developing some YAML check to add in to this script. I wouldn't really implement and deploy it, though, until #414 is finished and merged, since there may be some YAML changes because of that. I'll open a new to-do issue for that, but hold off work until the design changes are merged.

mdlincoln:
Flagging spanish language team for translation @arojascastro, @mariajoafana,
@vgayolrs

arojascastro:
Hello @mdlincoln  @jerielizabeth I am sorry, but I cannot figure out what changes need to be done in our guidelines for editors. I cannot find your changes here http://programminghistorian.org/editor-guidelines  

I know that I can see the changes you made in this pull request, but I would like to check the whole page first of all to find where they should go - it is not easy to find the exact sentence or section when dealing with translations because my departure point is the English text.

mdlincoln:
@arojascastro You are right, the "changes" view is limited. This link https://github.com/programminghistorian/jekyll/blob/c1636fd17fa5c3f44bffa70a84011e1db1afa4c9/editor-guidelines.md should show the entire file.

You can navigate to that link for any Pull Request by clicking on the "View" button in the "Files Changed" tab:

screen shot 2017-05-27 at 9 26 08 am

mdlincoln:
On hold until after redesign branch is merged, as there will need to be a few updates to the English text, and translation should happen on the new version of the Spanish editor guidelines page.

PR #439, opened 2017-05-20
Custom 404 page

mdlincoln:
Right now, if someone navigates to a bad link on our page, they get the default GitHub Pages 404 page:

<http://programminghistorian.org/nopagehere>

This PR creates a PH-specific 404 page, so that when a visitor hits a nonexistent link on our domain, they'll still see our banner, an apology, and a prompt to submit an issue if they think there's a problem.

screen shot 2017-05-20 at 4 17 46 pm

Most importantly, this page will also get tracked by Google Analytics, so we can monitor what URLs people are trying to use to get to our site, and add redirects as needed.

Does everyone like the banner image? Some alternatives include:

- https://www.flickr.com/photos/britishlibrary/11264668164/ (humorous)
- https://www.flickr.com/photos/britishlibrary/11223540985/ (extinction)
- https://www.flickr.com/photos/britishlibrary/11089149033/ (ruins)


ianmilligan1:
I _love_ it! Great picture choice, and great idea.

acrymble:
I like it, but the term 'issue' is jargon.

wcaleb:
Also might want to grayscale the image so it looks like the lesson gallery images and fits better with new design.

ianmilligan1:
Why not just "please let us know if you think there ought to be something here," with a link to http://programminghistorian.org/contribute. 

Although to be honest I don't think issue is too much jargon in this context.

mdlincoln:
@acrymble I think you are right that "issue" might be a bit jargony. However I think sending someone to the "Contribute" page would just be confusing. I got a bad link... and now I need to volunteer as a peer reviewer? I'll revise the language to something like "send us a message"

Also, for what it's worth, [no one will click on that link, anyway](https://blog.codinghorror.com/creating-user-friendly-404-pages/). The best use of the page will be collecting those bad links via Google Analytics, and helping fix those that we can.

PR #484, opened 2017-06-19
Add Sonic Word Clouds Blog Post

walshbr:
This blog post should be all ready to go, but I thought it made sense to wait until the design updates had been merged in just in case something wonky. I'll go ahead and merge in once those updates are squared away (and update the date of the post so that it's not backdated).

Also looks like my linter/editor made some unintended white space deletions to the bio file - the only entry that was actually modified was the one for Daniel Ruten, the author of the post.


mdlincoln (dismissed):
I don't think this should cause any challenges for the redesign, so I think you can go ahead and add a date to the post YAML and we can publish!

walshbr:
Will merge and publicize first thing in the morning.

PR #493, opened 2017-06-29
CC license translations

mdlincoln:
This modification would point to the EN version of the CC-BY license from EN pages, and the ES translation of the CC-BY license from ES pages. Does this sound acceptable?

More on the translations for CC licenses: https://creativecommons.org/faq/#what-are-the-official-translations-of-the-cc-licenses-and-cc0

### DIFF
--- a/_data/snippets.yml
+++ b/_data/snippets.yml
@@ -94,6 +94,9 @@ footer:
   license:
     en: license
     es: licencia
+  cc-link:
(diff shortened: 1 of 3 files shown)


acrymble:
Does this imply a legal change of license depending on what version you're reading? These licenses are about what laws apply in disputes.

mdlincoln:
No. The FAQ I link addresses that.

Issue #303, opened 2016-10-28
Project Team presentation - simplify?


acrymble:
I'd like to suggest that we simplify the Project Team page. There's too much text and I don't think it actually helps potential users to understand who to contact for what. It also asks people to contact Jeri Wieringa in the first instance, but she doesn't have an email address listed on the page.

I think the DH Commons masthead (equivalent to our project team) is more user-centric, and less about the editors.
![screen shot 2016-10-28 at 11 23 38](https://cloud.githubusercontent.com/assets/646647/19803562/572e0b5c-9d01-11e6-94a2-cca1af45a6c5.png)

We don't have to go that bare, but making it transparent who the editors are and how the project team comes together might be simpler for users. Categories such as:
- Commissioning Editor
- Editorial Team
- Spanish Language Team
- Community Development Team
- Technical Team

And to get your name on a respective list, you just have to be actively engaged in that aspect of the project?

The editor interests was a good idea, but I don't think it's translated into anything helpful for authors, and it's really just highlighted that we have narrow interests relative to the whole of digital humanities!

Thoughts?

ianmilligan1:
Sounds like a great idea to me, Adam. I think a more minimalist approach would be nice (although there might be something humanizing about having our smiling/stoic/etc. faces there).

fredgibbs:
assigning this to me, as per the general site CSS and layout update.

arojascastro:
I had a look at my bio and it is already outdated -- I am not longer based in Barcelona. Who I should contact to update it? Thanks.

acrymble:
@arojascastro you can update it yourself here: https://github.com/programminghistorian/jekyll/blob/gh-pages/_data/authors.yml

You raise an interesting question though about our records. Don't we want the record of where people were when they contributed (wrote, reviewed, whatever) rather than where they end up? I suppose in that sense someone could have multiple yml entries. But maybe we don't want to worry about that right now?

ianmilligan1:
I think that might make things overly complicated, having multiple yml entries. If we really want a static snapshot of where somebody was when they contributed a piece, it should perhaps just be hardcoded into the lesson?

acrymble:
We're a bit off topic now for this thread, but @ianmilligan1 you raise a good point. Not sure why we haven't been going with hard-coding.

acrymble:
[...]

PR #541, opened 2017-07-29
Editor roles

mdlincoln:
Closes: #303 

This is a test implementation of editor role badges and definitions.

Preview at: https://editor-roles--ph-dev.netlify.com/project-team

To-do:

- [x] Display `(Spanish)` qualifier for roles with editors, but don't have separate definitions.
- [x] Clicking on a badge should show you the definition (either by skipping down to the definition [easy] or showing a popup [a bit more tricky])
- [x] Spanish translations for terms and definitions needed
- [x] Adjust HTML/CSS so bios don't overflow


acrymble:
Looks great.

I'd like to cut the bios though. Just name and affiliation. The information readers need is captured by the tags.

ianmilligan1:
FWIW, I like knowing a brief bio from somebody before I contact them or work with them, gives me some context.

acrymble:
Opinions always worthwhile @ianmilligan1 but so that we can move forward can you be specific about what we (all) should have on our profile? I think consistency is key, and we discussed this at a call a few months ago and I believe we agreed that the editor interests left a lot of gaps, so weren't as useful as we might have intended.

At the moment the differing lengths of people's bios has caused some css issues as well (icons below pics).

ianmilligan1:
I think what we have right now is fine, with the addition of tagging. I guess we can put word limits if you think they're too long for CSS reasons.

But bios are one way for folks to be creative, express themselves, and let people know who they're dealing with.

acrymble:
If a personalised bio is important to people then the quickest solution is just to adjust the CSS. Otherwise the max is 2 lines per person.

It's also been pointed out to me that because the whole section is labelled 'Editorial Board', the tag for that is actually redundant.

ianmilligan1:
Whatever is easiest works for me - making folks switch to two-line bios or adjusting the CSS. But I do think a brief biographical statement is important.

mdlincoln:
I agree with @ianmilligan1 that I appreciate the bios tell me a bit more than just affiliation.

Regardless what we eventually stipulate for bios, however, I'd like to fix the CSS so that it still looks good _no matter how long_ the bio.

acrymble:
Minor point, but can we add a "Team Description" header (or something like that), to separate it from the people?

mdlincoln:
Thanks @vgayolrs for making those translations so quickly. I'll go ahead and merge this tomorrow - in the meantime, please let me know if there are any further comments or improvements we'd like to see.

Issue #393, opened 2017-03-22
Mallet lesson needs updating for more recent version


ianmilligan1:
@jburnford noted in an e-mail to me:

>The Programming Historian Mallet lesson seems to cause a small problem if students download the newest version (.8), instead of the .7 version you’ve got linked on the lesson. The final _composition.txt file does not work right. It is confusing as everything else seems to work until they open the file in Excel and it does not have the topic-percentage pairs for each file. The easy solution is to just get them to download the older version of Mallet, but if you and Shawn have a bit of time at some point in the future, it would be good to update it and figure out which part of the command changed in the new version.

We've made explicit that we should use the old version, but we really should update it at some point to use the newest version.

ianmilligan1:
I will do so when I have some spare cycles but might not be for a while.

acrymble:
I don't think you need to feel compelled to update this. The lesson is still internally correct. But it's up to you.

ianmilligan1:
Hah thanks Adam. We've had a fair number of questions about this - the version we use is no longer easy to find except through our lesson.

fredgibbs:
This sort of thing might be an opportunity to generate additional community participation--to see if we can't encourage someone who has been thinking about contributing to PH but doesn't want to write a new lesson--updating an existing one and replacing a few screen shots isn't too intimidating and could be a good trial run.

mdlincoln:
Flagging this for consideration re: #536

ianmilligan1:
Why don't us authors talk and see if we can get this updated.

mdlincoln:
https://github.com/programminghistorian/jekyll/issues/536#issuecomment-318408907

PR #544, opened 2017-07-31
Upgrade MALLET Lesson from 2.0.7 to 2.0.8

ianmilligan1:
Picking up on #393, this PR upgrades the lesson from 2.0.7 to 2.0.8. The major change in this release for our purposes is one of the output files, which can be seen in the new lines 524-532 below.

Putting here just to have it one place before merging. Any comments or thoughts are greatly appreciated.


shawngraham:
Given that 2.0.8 makes the output a bit less useable (from our standpoint), maybe just point the download to 2.0.7 & make a note that if the user grabs 2.0.8, the output is in this other format...?

ianmilligan1:
Re: keeping it at 2.0.7, that was our original plan, but it's leading to a lot of feedback... I think a lot of people download MALLET, get it started, and then turn to us. And one of the only places to find 2.0.7 is to dig into the link in our lesson.

We're trying to avoid lessons on outdated versions, which I think is probably a good policy.

acrymble:
You're both authors on this (as is Scott Weingart). This strikes me as the type of scenario we've been discussing a lot lately. @shawngraham it would be good to get your candid view on how you'd like this type of updating managed (assuming you have a candid view).

ianmilligan1:
Yes, we've also been discussing off this ticket.

mdlincoln:
No rush at all on this @ianmilligan1 - but do you have some target timeline for completing this? Just so I know when to check in 👍

ianmilligan1:
Just waiting on folks - my guess and hope is within two weeks of DH 2017 wrapping up. 😀

ianmilligan1:
Just to update re: this, I pinged @shawngraham and @scottbot.

PR #570, opened 2017-08-20
Remove manual lesson image step from editor guidelines

mdlincoln:
With our new lesson page infrastructure, editors no longer need to manually add lesson images and text to the lessons pages. This part of the editorial guidelines can be removed. 

Thanks @jerielizabeth for pointing this out - I think this is just a simple matter of removing that one paragraph? @vgayolrs I also removed the paragraph from the Spanish version. Let me know if either one seems to have problems.


jerielizabeth (approved):
Thanks @mdlincoln! That looks great to me!

vgayolrs:
@mdlincoln , that's ok

vgayolrs (commented):
OK to me!

PR #572, opened 2017-08-23
Exploring and Analyzing Network Data with Python

walshbr:
Going to do one last read through for copy and code checking before merging this new lesson in, but went through all the guidelines already. And then I'll advertise the piece next week so as not to conflict with the PR push for the OpenRefine lesson that went live this past week. Will soft launch, notify the authors to check it out, and then advertise next week.


mdlincoln:
FYI with our Netlify service now, you can see a preview of any PR by looking at the checks performed:

screen shot 2017-08-23 at 1 50 30 pm

https://deploy-preview-572--ph-dev.netlify.com/

walshbr:
@mdlincoln that's super nifty.

Issue #571, opened 2017-08-23
Editor Guidelines Outdated for Authors.yml


walshbr:
The example syntax for adding to authors.yml is outdated [here](https://programminghistorian.org/editor-guidelines#2-create-an-author-bio)

```
- name: Jim Clifford
   bio: |
       Jim Clifford is an assistant professor in the Department of History
       at the University of Saskatchewan.
```
Should likely be slightly more verbose based on the recent examples:
```
- name: Jim Clifford
   team: false
   bio: 
       en: |
             Jim Clifford is an assistant professor in the Department of History
             at the University of Saskatchewan.
```

PR #573, opened 2017-08-23
Update editorial guidelines refs #571

walshbr:
Update the editorial guidelines for adding authors to the ph-authors.yml file a la #571. Syntax seemed out of date given the examples that are currently in there - I think it was referencing older syntax from before the website redesign and Spanish translations. Flagging @arojascastro to update the Spanish guidelines accordingly and @mdlincoln just to make sure I'm on track here.

### DIFF
--- a/editor-guidelines.md
+++ b/editor-guidelines.md
@@ -182,9 +182,11 @@ If the lesson has been written by a new author, editors should add information a
 
 ```yaml
 - name: Jim Clifford
-  bio: |
-       Jim Clifford is an assistant professor in the Department of History
-       at the University of Saskatchewan.
+  team: false
+  bio:
+      en: |
+          Jim Clifford is an assistant professor in the Department of History
+          at the University of Saskatchewan.
 ```
 
 **Whitespace is important**, so be sure that the indentation matches the other examples.
--- a/editor-guidelines.md
+++ b/editor-guidelines.md
@@ -182,7 +182,6 @@ If the lesson has been written by a new author, editors should add information a
 
 ```yaml
 - name: Jim Clifford
-  team: false
   bio:
       en: |
           Jim Clifford is an assistant professor in the Department of History


vgayolrs:
@walshbr , I take this

vgayolrs (commented):
@walshbr In principle, I agree with your proposal. But there are also entries of authors with other metadata that I do not know if it is convenient to include in the editor's guide (in case we have this data):
```
- name: Jonathan Reeve
  url: "http://jonreeve.com"
  twitter: j0_0n
  github: JonathanReeve
  team: false
  bio:
    en: |
        Jonathan Reeve is a digital humanist and web developer for the Modern Language Association.
```
@mdlincoln , what do you think?

walshbr:
@vgayolrs Was wondering myself about that. My sense was that the extra metadata you've identified there was only used for those of us on the project team, so it might be alright to leave out of the template. But open to other opinions.

mdlincoln (changes requested):
It's not necessary to explicitly state `team: false`, as Liquid parses the lack of a `team` key as `team: false` by default.

mdlincoln:
And yes, the additional metadata on some author entries is only used for displaying editorial team status - so it certainly doesn't need to be added.. If `team: false` was the only change, then I think... this PR doesn't need to be merged?

If we want to be completely clear, you could add a sentence instead saying "just use the bare minimum bio for authors; extra metadata seen in other entries is only for editorial team members"

walshbr:
Cool. @mdlincoln the only other bit I added was "en" where it wasn't already there - is the metadata also parsing as English by default?

vgayolrs:
Thanks for the explanation, @mdlincoln . So, it remains the same.

mdlincoln:
@walshbr oh crap, how did I miss that addition? Yes, it _does_ need the `en` subkey.

editor-guidelines.md
```diff
@@ -182,9 +182,11 @@ If the lesson has been written by a new author, editors should add information a
 
 ```yaml
 - name: Jim Clifford
-  bio: |
-       Jim Clifford is an assistant professor in the Department of History
-       at the University of Saskatchewan.
+  team: false
```
mdlincoln: Remove this line only. Rest is good.

mdlincoln (dismissed):

walshbr:
Cool! Should be good to merge now - just needs the matching Spanish bit.

Also - this is fantastic documentation. So thorough and clear.

vgayolrs:
@walshbr , done!

PR #577, opened 2017-08-24
Fix for api filtering

### LINKED ISSUE #574: Sort by API Counting Incorrectly
Sorting by topic for API on the lessons page is counting the number of lessons on the page incorrectly. It lists 1 as the number of lessons for that topic, but selecting it produces 6 lessons in the sorting interface.

walshbr:
@mdlincoln can you glance over the search filtering here to just make sure I haven't gummed up the works? Should fix #574. Looked like it was loading topic categories all as one big string - so "webscrapingapiwebsite" for example. Then when filtering it was just just looking for a substring to match. Because "webscraping" has the letters "api" in it the filtering was getting confused. Switched it to add spaces between the topic categories. Then, when looking against those topic categories, it now splits them apart into an array on spaces and searches for a matching item in the array.

### DIFF
--- a/_includes/lesson_describe.html
+++ b/_includes/lesson_describe.html
@@ -32,7 +32,7 @@ lesson-slug.html computes the correct lesson slug, and then this include determi
  Activity and topic are hidden via CSS as a hack to let JS sorting by activity and topic work
 {% endcomment %}
   <span class="activity">{{ lesson.activity }}</span>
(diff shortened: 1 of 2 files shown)


mdlincoln (approved):
javascript whyyyyyy 

hashtag javascript

ianmilligan1:
+1 to the GIF

PR #583, opened 2017-08-30
Baseurl to URL and add HTTPs to Config

### LINKED ISSUE #582: Lesson Titles Resolving to Broken URLs
@shawngraham reported this to me.

Right now, if you click on any of the lesson titles – i.e. "Web Mapping with Python and Leaflet" in the below screenshot, or any other lesson, it resolves to a broken URL: i.e. `https://lessons/mapping-with-python-leaflet`

screen shot 2017-08-30 at 9 31 11 am

or

screen shot 2017-08-30 at 9 32 15 am

Found on https://programminghistorian.org/lessons/mapping-with-python-leaflet, https://programminghistorian.org/lessons/exploring-and-analyzing-network-data-with-python, and https://programminghistorian.org/lessons/fetch-and-parse-data-with-openrefine (and presumably more, those were just the first three I checked).

walshbr:
Two changes -

Layout for lessons was pulling in baseurl, which is an empty string at the moment. Using the url variable instead. Also noticed that the url in the config file was still http and not https, so I updated that to use https as well. Should resolve #582

### DIFF
--- a/_layouts/lesson.html
+++ b/_layouts/lesson.html
@@ -28,7 +28,7 @@ layout: base
 
       <div class="col-md-8">
         <div class="header-title">
-          <h1><a href="{{ site.baseurl }}/{{ page.url }}">{{ page.title }}</a></h1>
+          <h1><a href="{{ site.url }}{{ page.url }}">{{ page.title }}</a></h1>
         </div>
 
         <div class="header-author">
--- a/_layouts/lesson.html
+++ b/_layouts/lesson.html
@@ -28,7 +28,7 @@ layout: base
 
       <div class="col-md-8">
         <div class="header-title">
-          <h1><a href="{{ site.url }}{{ page.url }}">{{ page.title }}</a></h1>
+          <h1><a href="{{ site.baseurl }}{{ page.url }}">{{ page.title }}</a></h1>
         </div>
 
         <div class="header-author">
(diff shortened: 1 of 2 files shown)


walshbr:
Might be the case that you'd have that be a relative rather than absolute URL, so just let me know.

_layouts/lesson.html
```diff
@@ -28,7 +28,7 @@
 
       <div class="col-md-8">
         <div class="header-title">
-          <h1><a href="{{ site.baseurl }}/{{ page.url }}">{{ page.title }}</a></h1>
+          <h1><a href="{{ site.url }}{{ page.url }}">{{ page.title }}</a></h1>
```
mdlincoln: I'd prefer this to be a relative URL instead of absolute. It's odd that `site.baseurl` doesn't map correctly... but anyway, can you change this to:

``` html
<h1><a href="/{{ page.url }}">{{ page.title }}</a></h1>
```

mdlincoln (dismissed):

walshbr:
@mdlincoln looks like the problem was in the '/' character actually - seems like it was automatically adding a '/' and getting rid of that lets us use baseurl. That actually seems to be what the intended behavior was, so reinserted baseurl but took out the offending character. Let me know if you'd still rather this be 

```
<h1><a href="/{{ page.url }}">{{ page.title }}</a></h1>
```
 instead.

PR #606, opened 2017-09-18
Staging for the Blog: post on CLiC and corpus analysis

jerielizabeth:
Hi all, 

I'm staging a blog post that has been in the works this summer. It is not quite ready to be merged in yet, but I needed the preview feature for the authors. The tools itself was too focused for an independent lesson, but we thought a post would be a good way to build on the AntConc lesson and show some applied corpus linguistic analysis.

As always, any editorial suggestions are welcome!


jerielizabeth:
I see what is going on - using the `include figure` syntax, the caption is placed inside a `<p>` tag, which overrides the figcaption display setting and adds the bottom margin. I'll switch most of the images here to that format. But I think some css changes to the figcaption tag would also be useful, and will suggest those via a pull request.

PR #625, opened 2017-10-15
New snippets for Spanish lesson index

### LINKED ISSUE #623: Snipets of Spanish new lesson page must be translated
@mdlincoln : there's a little detail on the new Spanish lesson page, which you could possibly help me sort out: https://programminghistorian.org/es/lecciones/

When I choose a filter for the types of lessons (by date or by difficulty or by topic), some of that information appears in English and some in Spanish.

For example, "filtering results" should be translated to "resultados de filtrado", "all lessons" to "todas las lecciones", "date" to "fecha" and "difficulty" to "dificultad", and so on. When I filter by `data management`, I get the correct sentence in Spanish (`gestión de datos`).

I tried to find if there is a code file to do it, but I only found the animated gifs in this folder: https://github.com/programminghistorian/jekyll/tree/gh-pages/images/new-lessons-page

If you tell me where I can go to translate these fragments, I will thank you very much.

mdlincoln:
~While waiting to figure out the JS problems on #623, we can at least translate these snippets.~

Closes #623


mdlincoln:
@vgayolrs can you please push commits here: https://github.com/programminghistorian/jekyll/blob/lesson-filter-es-fix/_data/snippets.yml#L135-L140

Please add your commits to the `lesson-filter-es-fix` branch

vgayolrs:
Thanks @mdlincoln . I'm on it. Let me review carefully some details. I'll let you know when I'm done.

vgayolrs:
ok @mdlincoln , I think it's ready to merge.

vgayolrs (dismissed):

mdlincoln:
Thank you @vgayolrs. We still need to fix the Javascript issue before these snippets can be displayed, unfortunately. I'll keep working on it.

vgayolrs:
You're welcome @mdlincoln . Let me know if there's anything else I can do. Unfortunately, no javascripts at the moment.

mdlincoln:
So @walshbr I know _no javascript whatsoever_ but I think I hacked my way through this. As described in the original issue, the problem is that we can't use liquid includes within lessonfilter.js, so we need to find another way to pass the proper text labels through to javascript. I think I've done that here.

One question: do you know how to properly set `$('#date-sort-text').attr('label')` and `$('#difficulty-sort-text').attr('label')` into variables so they properly render in that `text()` function? I feel like this could be DRY-er - right now I call each of those queries twice, which is awkward.

walshbr:
That what you had in mind @mdlincoln? Check out the last commit where I set the variables at the bottom of the js file. One issue with how I did this, I think, is that it sets the labels on each page load. So if the file gets loaded on a page where those labels don't exist we'd get some unexpected behavior. But it appears that this js file is only getting included on the lessons page anyway, so that problem never arises.

And great job figuring it out! I couldn't quite wrap my head around what you meant in the original note on this issue, so I'm glad that you figured it out.

mdlincoln:
Ah I see - I'd tried saving them to variables somewhere else in the JS and it wasn't working.

I think this is ready for merging if you agree @walshbr

walshbr (approved):
Looks good to me! We can open separate tickets if anything arises. But seems like the functionality is working as expected, and I'm not seeing errors in the console anywhere.

Issue #629, opened 2017-10-17
Basic Text Processing in R: Word Frequency


ndalyrose:
Several of us working through this section of the tutorial on word frequency received this error and couldn't figure out how to resolve. We're on Windows machines.:

> wf <- read_csv(sprintf("%s/%s", base_url, "word_frequency.csv"))
wf
Error in open.connection(con, "rb") : 
  Failed to connect to programminghistorian.org port 80: Timed out

acrymble:
do you know what it's supposed to be finding? I didn't work on this lesson, but it looks like it's trying to call a file somewhere. Maybe the URL is just wrong?

acrymble:
https://programminghistorian.org/assets/basic-text-processing-in-r/word_frequency.csv this looks like the correct URL. I suspect there's an issue with "%s/%s" not creating that from the base_url...

drjwbaker:
I *think* the line should be `wf <- read.csv(sprintf("https://programminghistorian.org/assets/basic-text-processing-in-r/word_frequency.csv"))`, but @statsmaths et al are the experts.

jerielizabeth:
I think the problem is with the base_url value -- the lesson was pre our switch to https. I tried it locally and got an error using http in the base_url value, and no errors when switching to https. @ndalyrose, could you try changing the base_url value from `base_url <- "http://programminghistorian.org/assets/basic-text-processing-in-r"` to `base_url <- "https://programminghistorian.org/assets/basic-text-processing-in-r"` (the first steps in the exploratory analysis section - https://programminghistorian.org/lessons/basic-text-processing-in-r#exploratory-analysis) and see if that resolves the issue?

jerielizabeth:
If that works, I can push up a pull request with the change.

taylor-arnold:
My guess is that @jerielizabeth is correct about it having to do with http versus https. Both the original and the https versions worked fine on my computer though, so I am not a good judge of whether this fixes the original issue (could be that my machine is switching over automatically due to my setup). @ndalyrose, does the https change fix it for you ?

jerielizabeth:
I've queued up the change, but will leave the issue open for feedback from @ndalyrose et al.

jerielizabeth:
logging this in case https does not fix the issue. If I remember correctly, @walshbr was a reviewer on this issue and did run into some intermittent trouble with network connectivity on a university network.

walshbr:
[...]

PR #632, opened 2017-10-19
shift base_url value to https. should address #629

jerielizabeth:
This (very small) change to the base_url value should address the error in #629.


mdlincoln:
aaaghh, `base_url` strikes again!

This is a good change regardless - but yes, I want to hear from @ndalyrose if using https fixes her issue - as, like @statsmaths, I can't reproduce the issue on my machine.

mdlincoln (approved):

PR #645, opened 2017-11-02
Adding Anouk Lang to peer reviewer list

ianmilligan1:
I forgot to add Anouk to `reviewers.yml`. Should be an easy merge, and I do not believe it requires translation. We also didn't have Lee Skallerup Bessette.


mdlincoln (approved):

mdlincoln:
A good reminder that I never added reviewers for my last lesson on there! Thanks.

PR #652, opened 2017-11-13
upload translation intro to lod

arojascastro:
Please @vgayolrs @mariajoafana @mdlincoln review and accept the upload of a new translation.


arojascastro:
How many reviewers do we need for this? And who should be they?

ianmilligan1:
I would say in the case of a pull request where it's a new Spanish lesson being uploaded to go live, one of @arojascastro @vgayolrs @mariajoafana could review and approve? (thoughts @mdlincoln?)

At least when the checks pass. In this case, there is a build failure (clicking on Details shows):

```
* In /es/lecciones/intro-a-datos-enlazados.md:
  - [ ] 'activity' is missing.
  - [ ] 'topics' is missing.
  - [ ] 'abstract' is missing.
  - [ ] 'translation-reviewer' is missing.
  - [ ] 'translation_date' is missing.
jekyll 3.5.2 | Error:  There were YAML errors.
```

The advantage of using pull requests like this is that we can fix all the problems on a branch before going live. In the past for example I've put something up with broken links (oops!), and then we wait for Travis and fix it, etc. so this makes it easier for all of us. 😄

vgayolrs:
An apology for the question, @arojascastro , but, a new translation that has not gone through translation review, should not be uploaded in ph-submisions first?

arojascastro:
@vgayolrs yes, you are right, sorry! I put it in a different directory... so I'll undo this an try to push it in the correct place. Thanks to all

Issue #637, opened 2017-10-25
Missing "contact information" on Project Scholarship page


ianmilligan1:
Right now, the pre-amble to the "Project Scholarship" page reads like this:

>The project team and members of the wider community are involved in a number of scholarly initiatives related to our work here at The Programming Historian. These include events, journal articles, reviews (of us by the community), and posters. If you are conducting academic research using this project please let us know.

It says "please let us know," but doesn't provide a contact person. We should add an e-mail for the managing editor, as is standard throughout the site.

_NOTE: This is part of a test issue/pull-request sequence to support project documentation._

PR #655, opened 2017-11-17
translate update #637

vgayolrs:


ianmilligan1 (approved):
Looks great! In the future, we should probably try to keep both English and Spanish on the same pull request, but let's move this forward. Thanks so much @vgayolrs!

PR #663, opened 2017-11-30
Add archivist role to teamroles and snippets

drjwbaker:
add archivist role. Is this the right wording Adam?

needs translation

PR #664, opened 2017-11-30
Update ph_authors.yml

drjwbaker:
Add role per https://github.com/programminghistorian/jekyll/pull/663


mdlincoln:
Please put this into the same PR as the changes to the snippets file in #663 , since we need to test and preview them in tandem.

PR #672, opened 2017-12-20
Fixes "next lesson" on series of lessons

### LINKED ISSUE #671: Navigation to next lesson is broken
At the end of each lesson the links to advance seem to point to the previous lesson instead.
"Great! Now you're ready to move on to the next lesson ."
So this text at the end of lesson 2 points back to lesson 1 instead of pointing to lesson 3

All the chapters seem to be like this.
These are the lessons on Python that should begin at:
https://programminghistorian.org/lessons/introduction-and-installation
You will also see that this first lesson has no way to get to the second lesson.

ianmilligan1:
The "next lesson" instructions in the lesson template was erroneously linking to the previous lesson in the series, not the next one. Tested locally and it now works properly.

This will resolve #671 when merged.

### DIFF
--- a/_layouts/lesson.html
+++ b/_layouts/lesson.html
@@ -154,7 +154,7 @@ layout: base
 
      {% if page.next %}
       <div class="series-warning alert alert-info">
-        {{ site.data.snippets.series-next-note[page.lang] }} <a href="{{page.previous}}">{{ site.data.snippets.next-lesson[page.lang] }} </a>.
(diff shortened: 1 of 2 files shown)


walshbr (approved):

ianmilligan1:
Do you want to hit merge @walshbr if it all checks out on your end?

walshbr:
Ah sorry - wasn't sure if the reviewer was also supposed to do  that.

ianmilligan1:
Heh maybe worth chatting. I think it's worth just getting into the practice of never merging our own pull requests, however trivial they might be.

PR #694, opened 2018-01-25
downsize MJAL avatar

mdlincoln:
I'd noticed that Maria's avatar was loading in quite slowly - turns out it was a very large image file. This PR downsizes it so our team page will load faster.


ianmilligan1 (approved):
lgtm!

PR #699, opened 2018-01-25
Fix dates

### LINKED ISSUE #697: "Last updated" header showing nonsensical number
The latest update to the [Sentiment Analysis lesson](https://programminghistorian.org/lessons/sentiment-analysis) (#692) is causing our "last modified" header to show the nonsensical date `2018-00-25`

I need to look in to what part of the jekyll/js code may be causing this.

walshbr:
Should close #697 with thanks to @zoeleblanc for the syntax help!

### DIFF
--- a/_layouts/lesson.html
+++ b/_layouts/lesson.html
@@ -169,7 +169,9 @@ layout: base
   })
     .done(function( data ) {
       var date = new Date(data[0].commit.author.date);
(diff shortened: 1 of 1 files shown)


mdlincoln:
I.... would sort of like to fix it with a more solid option that doesn't splice strings around. See #700

PR #700, opened 2018-01-25
Use core JS to parse dates from GH api

mdlincoln:
Closes #697

PR #701, opened 2018-01-26
Fixing a bug with the month of last modification

symac:
date.getMonth() returns a result [ranging from 0 to 11](https://www.w3schools.com/jsref/jsref_getmonth.asp), we need to add 1 to it, otherwise the display is wrong. For January update : 00 instead of 01.

See an example of the bug on [Sentiment Analysis for Exploratory Data Analysis ](https://programminghistorian.org/lessons/sentiment-analysis) lesson that currently display "modified 2018-00-25". This patch fixes the issue and does not seem to break anything.

### DIFF
--- a/_layouts/lesson.html
+++ b/_layouts/lesson.html
@@ -169,7 +169,7 @@ layout: base
   })
     .done(function( data ) {
       var date = new Date(data[0].commit.author.date);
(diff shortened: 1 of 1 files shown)


mdlincoln:
This was fixed in #700

PR #612, opened 2017-09-29
Lesson retirement policy

mdlincoln:
This PR will show a preview of the Lesson Retirement Policy page draft re: #536 

Discussion of exact language and tasks will take place here. Any additions can be pushed to the `archiving-policy` branch.

Preview: https://deploy-preview-612--ph-dev.netlify.com/lesson-retirement-policy

Preview lesson: https://deploy-preview-612--ph-dev.netlify.com/lessons/retired/counting-frequencies-from-zotero-items

To-do:

- [x] reinstate `translation-editor` field checking before merging, temporarily removed so site will properly build (re: #611) ~(do by reverting 318b712bc23c987b48d0490e65146636c5bdb6d2)~ **already completed and merged to `gh-pages`**
- [x] rename `lessons/deprecated` to `lessons/retired`, add appropriate redirects
- [x] Language about twitter bot #514
- [x] Mention sustainability goals #534 
- [x] Spanish translations
    - [x] New page on retirement policy 
    - [x] a few new lines in [_data/snippets.yml](https://github.com/programminghistorian/jekyll/pull/612/files#diff-a407e023d60dfefccac66d5a7959003c)
    - [x] additions to [author guidelines](https://github.com/programminghistorian/jekyll/pull/612/files#diff-72a7d8f7b711a8af34fe8532ab83c3a6)
    - [x] additions to [editor guidelines](https://github.com/programminghistorian/jekyll/pull/612/files#diff-542dbfd70101e87b830b47847632b7c1)
    - [x] additions to [reviewer guidelines](https://github.com/programminghistorian/jekyll/pull/612/files#diff-a038c8ad532ac4704b06f260e545d023)
- [x] List retired lessons
- [x] Fix display of images on retired lessons
- [x] add `retired_date` to retired lessons

PR #707, opened 2018-01-29
translate snippet and fix equipo text 

arojascastro:
sorry for creating a new branch for this... may you approve this please?

### DIFF
--- a/_data/snippets.yml
+++ b/_data/snippets.yml
@@ -219,4 +219,4 @@ project-team-roles:
 
 present:
   en: Present
-  es: Present
+  es: Presente
--- a/es/equipo-de-proyecto.md
+++ b/es/equipo-de-proyecto.md
@@ -26,9 +26,9 @@ All editorial board information should be edited in data/ph_authors.yml. Authors
 
 Agradecemos especialmente a los autores que han contribuido a _Programming Historian_ y que no son miembros del equipo del proyecto:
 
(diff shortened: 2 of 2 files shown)


mdlincoln:
It's ok! Once a PR is merged, we have to create a new branch anyway - and this time, it looks like it's just the changes we need. Thanks :)

mdlincoln (approved):

arojascastro:
Yes, but I should use an already existing branch, right? like *translation-es* rather than creating new branches every time I make a change, right?

mdlincoln:
If changed English text is part of a PR (like #612) then you should add the commits to the branch of that PR. You can see the name of a PR's branch right underneath its header. The PR #612 is on the `archiving-policy` branch:

screen shot 2018-01-29 at 8 46 24 am

If you are changing text that is already in `gh-pages`, then you need to create a new branch. There is no special `translation-es` branch that we keep updating, though.

arojascastro:
Ok, yes, the problem is that I am using SourceTree and sometimes I am not in the *right* branch, and I committ changes from the wrong branch... 

I have a *translation-es* branch... but maybe I should not?

captura de pantalla 2018-01-29 a las 17 52 39

mdlincoln:
Branches are temporary things - so we create new ones for changes that need to go together. The archiving policy changes should all be on the same branch for testing and review - but once they're merged, then we can delete that branch. The only branch that is really permanent is `gh-pages` which is the published version of the site.

Issue #499, opened 2017-06-30
Blog post on scoping lessons


acrymble:
Arising from our #475 call was a notion that new authors tend to come to us with overly ambitious scopes for potential lessons, and that they under-estimate the amount of time and effort that goes into a good lesson.

Rather than updating documentation, would people value a blog post that we could refer people to that gives some perspectives on the above problems, as a means of reducing editorial backlog and lessons that fail to materialise?

walshbr:
That does seem useful. While each case is different, some boilerplate we can refer people to might be a helpful time saver.

acrymble:
Proposed text for comment:

---
title: Tips for (finishing) a Good Technical Tutorial
authors: 
- Adam Crymble
layout: post
categories: posts 
---

So you've got an idea for a tutorial, and you're excited to share it with the readers of the _Programming Historian_. This post outlines a few tips for how to sculpt that idea into a manageable and sustainable lesson, while also giving a sense of the level of work involved. We hope that this will help our authors at those early stages of the process, and encourage them to bring project ideas that will result in finished lessons that our readers can benefit from.

## Sustainability

We do our best to future-proof our lessons so that readers can benefit from it for many years. You can help with this process by thinking ahead. Some ideas include:

1) Link to Wikipedia for any technical terminology and jargon. The folks at Wikipedia will keep the definitions up to date for you.

2) Try to pitch lessons that are not overly dependent upon specific software or user interfaces. These lessons inevitably break or need substantial revision when a new version comes out. Teaching concepts rather than 'click on the X button' helps make for sustainable tutorials.

3) Make sure any data required to do the lesson is included as an attachment, stored on our site. This ensures lessons do not break in future when links begin to rot.


## A Manageable Scope

We get a lot of lesson proposals, and the thing that unites most of them is ambition. Many of the pitches we receive simply want to do too much: teach an entire programming language, or cover everything one needs to know about a category of research. We love your ambition, but we also know from experience that you've just pitched a series of textbooks, and it isn't feasible. A lesson is best when it seeks to do one thing well. Take a look at our [published lessons](http://programminghistorian.org/lessons/) for a sense of what's possible.

## Multiple Lessons

[...]

PR #722, opened 2018-02-20
blogpost connected to #499

acrymble:
This is the blog post from issue #499, which we can point authors to when they are trying to come up with a manageable scope for their lessons.

### DIFF
--- a/_posts/2018-02-20-planning-a-lesson.md
+++ b/_posts/2018-02-20-planning-a-lesson.md
@@ -1,0 +1,48 @@
+---
+title: Tips for (finishing) a Good Technical Tutorial
+authors: 
+- Adam Crymble
+layout: post
+categories: posts 
+---
+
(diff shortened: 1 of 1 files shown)


mdlincoln (changes requested):
The core points are good here, but I think you might do another editing pass with an eye towards tone. Since this is supposed to be a useful post for cultivating would-be authors, I think we should avoid sounding exasperated, or like we are presenting a list of complaints (even though these are all very real challenges that you list!)

Perhaps orienting it a bit more towards the author's point of view, rather than that of an irritated editor, might help. For example, right now the post laments authors "disappearing" or needing to be chased down. You might reframe it as: we want to make sure that authors' initial hard work submitting a draft doesn't go to waste. By understanding the time commitment involved, authors can better focus on managing their scope and expectations about time commitment.

walshbr (approved):
I largely agree with @mdlincoln. This is all great advice, and I think just framing it in the ways he suggest might help. I'll go ahead and approve since the changes I would request are the same! Thanks for putting this together, @acrymble. Really glad it will be on the blog.

acrymble:
Thanks, I've revised the tone. Good suggestion.

mdlincoln (commented):
This looks excellent! I'll merge it now.

mdlincoln (approved):

Issue #723, opened 2018-02-21
Menu is not displayed on mobile devices?


arojascastro:
I accessed to the PH on my iPhone/Firefox and I found out that the Menu tabs “About” and “Contribute” are not working properly. If you click on them, the sub-contents are not displayed. 

Can anyone else check?

walshbr:
Can confirm it's not working using my phone. You can't reproduce the issue in developer tools/web inspector on a desktop if you're simulating the mobile/responsive version of the site - more info [here](https://mobile1st.com/why-chrome-devtools-is-inaccurate-for-mobile-testing/). To actually work on the issue you'd need to plug the phone into a computer - more info [here](https://forums.tumult.com/t/debug-issues-on-ios-devices-using-safari-on-a-desktop/5180). I don't have the correct cable to connect my phone to my laptop at work today, but I can take a look later in the week if other folks can't get to it sooner. The submenus are likely there and just being hidden by the CSS.

ianmilligan1:
Confirmed that it's not showing up on my iPhone either.

mdlincoln:
*curses in CSS*

mdlincoln:
Do you feel able to take point on this @walshbr ? I'm glad to review what you do, but CSS is... not my forte :(

walshbr:
Yeah I can take a look @mdlincoln.

walshbr:
Looks like this is related to an underlying issue - we are running Bootstrap 4 alpha for css and js, which is a bit concerning since it's an alpha release. Bootstrap is on version 4 now, so we might want to migrate over to the stable release at some point. I tried that migration out just to see how complicated it was, and it looks like BS has changed enough that it would break some non-trivial CSS. That migration would need to be its own issue I think.

That's all to say that this dropdown CSS error is documented for the alpha release - the show class isn't getting add to the dropdowns on a click on mobile. I've been playing around with the hacks that others had to get it working, but no dice so far. I'll keep poking, but might have to just go ahead and migrate us to the stable release to fix it. I'll keep investigating.

walshbr:
[...]

PR #728, opened 2018-02-23
Mobile CSS Fix

walshbr:
Please don't review yet - am using the netlify preview for mobile testing.


mdlincoln:
re: #723

walshbr:
@mdlincoln or anyone else - this should be good to go now. I have two questions though: 

1) looks like i’ve been working so long that a few urls have gone down on my mobile-css-fix branch. i feel like i should open an issue for those separately, but does that mean i should just leave this branch unmerged because of those unrelated issues that are causing failing tests?

2) am i right in thinking that rebasing will collapse the number of commits that i have here into one? b/c i had to test with netlify i have so many garbage commits there that would be good to collapse to protect the git history.

mdlincoln (approved):

mdlincoln:
We can "squash" the commits on GitHub when merging, so don't worry about anything fancy on your end.

As for the 404s, let's address those separately if they still show up after we merge this one.

PR #740, opened 2018-03-13
New translation: Intro google maps

mariajoafana:
Live preview: https://deploy-preview-740--ph-dev.netlify.com/es/lecciones/intro-a-google-maps-y-google-earth

PR #750, opened 2018-03-20
split author bios into individual paragraphs

mdlincoln:
I'd noticed that the current layout of author bios looks a bit weird when a lesson has more than one author, with awkward spacing between each sentence:

screen shot 2018-03-20 at 8 52 18 am

This PR instead splits multiple bios into their own paragraphs like so:

screen shot 2018-03-20 at 8 54 41 am

Thoughts/reactions?


mdlincoln:
FYI the broken links are being fixed in #740

ianmilligan1:
I think this looks much better, @mdlincoln!

ianmilligan1 (approved):
LGTM!

PR #761, opened 2018-03-27
Adds asset file; link from lesson. Closes #758.

### LINKED ISSUE #758: Files not in assets directory
I am currently translating this tutorial on [XML and XSL](https://programminghistorian.org/lessons/transforming-xml-with-xsl). The necessary files to complete the tutorial are stored in the author's github repository. I was wondering if we should store them in ours instead. See this section: https://programminghistorian.org/lessons/transforming-xml-with-xsl#choosing-and-preparing-xml-data

On the other hand, I have a language doubt: the author writes *aggrevation* in the introduction:

> In all three of these situations, a basic understanding of XML, and its sister-language XSL, could have saved you time and aggrevation. This tutorial will provide you with the ability to convert or transform historical data from an XML database (whether a single file or several linked documents) into a variety of different presentations—condensed tables, exhaustive lists or paragraphed narratives—and file formats. Whether filtering a large database or adding formatting such as headings and pagination, XSL offers historians the ability to reshape databases to support their changing research or publication needs.

Is it correct? Or it should be *aggravation*? I looked for that word in the dictionary and found nothing.

ianmilligan1:
This would resolve #758. It does the following:
* Adds the zip file of the assets directly;
* Adds a link in the lesson. I think the idea of leaving the original instructions makes sense and respects the authors' intentions (plus it's good for people to see that they can download entire GitHub repositories as a ZIP).

### DIFF
--- a/lessons/transforming-xml-with-xsl.md
+++ b/lessons/transforming-xml-with-xsl.md
@@ -119,6 +119,8 @@ To begin work with the Scissors and Paste Database, visit its [Github repository
 
 {% include figure.html filename="transforming-xml-with-xsl-1.png" caption="Figure 1: Downloading Your Data" %}
 
+Alternatively, you can also download the scissors and paste file [here](../assets/scissorsandpaste-master.zip).
+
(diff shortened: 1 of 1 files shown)


acrymble (approved):
this change is fine, and improves the sustainability of the lesson.

PR #771, opened 2018-03-30
Suggested Citations

mdlincoln:
Closes #767

Thoughts?

Example lesson in english: https://deploy-preview-771--ph-dev.netlify.com/lessons/text-mining-with-extracted-features

Example lesson in spanish: https://deploy-preview-771--ph-dev.netlify.com/es/lecciones/escritura-sostenible-usando-pandoc-y-markdown

PR #774, opened 2018-04-02
use snippets to implement language-specific title quoting

mdlincoln:
re: #771

I broke this out into snippets because if we implement yet more languages, we'll need to deal with the possibility of using e.g. `«»`


arojascastro (approved):
Wonderful and clever

PR #804, opened 2018-04-18
correct wrong translation date

arojascastro:
it couldnt be translated before being published in English


mdlincoln:
😂 haha I'd run in to this lesson once before, and got somewhat confused, but then I forgot to open an issue! Thanks for fixing.

mdlincoln (approved):

Issue #785, opened 2018-04-06
Refactor the way we point to lessons


mdlincoln:
Coming out of #784:

>On the technical side, I think this suggests a much larger refactoring job in the way that we establish "identity" for lessons. Right now the workarounds like redirect_from and so forth rely on the core assumptions that we are always translating from English into Spanish. It uses a lot of string interpolation and guessing. This refactoring instead locates the actual target lesson and draws directly on its metadata, which ought to ensure more correctness. It'd also somewhat future-proof us, as it'd easily allow for spanish lessons to be originals and english ones to be translations, to add French translations, etc...
>
>That larger refactor to implement this pattern across the site is another job (and will need to be completed to accomplish #685) but this fix has helped me figure out a path forward that I think will also clarify the code-base a fair amount.

This is going to be a placeholder for a bit while I collect instances where our code can be updated to be more language-agnostic.

PR #813, opened 2018-04-23
Set titles of es about & index pages to Spanish

mdlincoln:
While working a bit on #785 I discovered that the "title" attributes for the Spanish versions of the home page and about pages were still in English.


arojascastro:
Muchas gracias :)

arojascastro (approved):

PR #816, opened 2018-04-27
Check for presence of `review-ticket` in lesson YAML

mdlincoln:
Don't merge yet! I'm using this PR as a way to run a check of all the lessons that don't have a `review-ticket` entry in their YAML metadata re #812

Issue #651, opened 2017-11-06
Language Independence


ianmilligan1:
I wanted to move the conversation from #647 so policy discussions don't get wrapped up with an individual lesson. 

@arojascastro noted below:

>I may be wrong but the point is that Geoparser only recognizes named entities in English; I have never used it but I checked the website and I did not see any way to change the language parameter or whatever. So if anyone translates this lesson, s/he will have to use the text in English - because it will not work with texts in Spanish.
>
>What do you think? @vgayolrs @mariajoafana
>
>On the other hand, maybe it would be good pracrice to accept the publication of lessons that takes into consideration how dependant the content is on language.
>
>Examples:
>
>* is it possible to use texts in different language other than English? Then it may be accepted
>* can you use resources that are transnational (for instance Europeana)? Fine!
>* are you using tools that are available in different languages? Perfect!
>
>In my opinion, if the PH is committed to open access and free tecnologies, it may be good idea to extend its committment to multilingual tecnologies as well... of course a first step could be tecnologies that are localisation friendly...

ianmilligan1:
My own two cents is that I think it's worth making a consideration, but I don't think we want to have yet another blanket policy.

The Geoparser is a good example: by default it focuses on English, but a research team could tweak different elements to make it work in different languages. It would take effort and time, and it's not realistically feasible for tool projects to implement all of this out of the box on the resources that most DH projects have. 

For what it's worth, my work for example focuses on English-language tools out of the box, mostly because (a) that's our expertise; (b) our pipeline of tools; and (c) our local users. We make it so that it's extendable down the road for other linguistic communities, but in some cases that's going to require people to do a bit of their own legwork on the text classifier front. With these complex tools, language independence isn't something you're going to easily have out of the box, especially given interdependent processing steps.

arojascastro:
Yup, I agree mostly, I think you could take into account when reviewing proposals - just as one factor amongst others. 

However, your scenario applies well to programming languages, tools and methods mostly. 

What about the reuse of information already published by third parties? I think a great amount of work can be on this second direction: to use resources that are transnational and multilingual. They already exist. So for instance, instead of mining the Internet Archive, why not mining Europeana? 

At the moment all PH contents, examples and resources are focused on the English culture, which of course makes total sense because authors are English or American, but the audience... is it English or American only? 

[...]

PR #818, opened 2018-04-27
Language links refactor

mdlincoln:
This is a major back-end modification that adjusts how to point to and from translations and their sources. As a result, we should now theoretically be able to host any number of languages on the site. Another nice benefit is that clicking on the `[es]` or `[en]` links at the top of the page should now - when there is a translation / translation source - take you to the companion page for the one you were just on, rather than dumping you all the way back at the home page for that language.

This works by doing two searches every time a page is generated: 

1. if the page has YAML metadata pointing to a source translation, this identifies that exact page and makes that source's metadata available when rendering the translation's page.
2. Searches all other pages across the site to see if they're _translations of the page being rendered_. If so, makes all the metadata of those pages available as needed.

This results in only one change for the editorial guidelines for producing translations: instead of using `redirect-from: /lessons/LESSON-SLUG`, editors will now use `original: LESSON-SLUG` (note no leading directory information). I've already gone through and updated all the existing translations for this. All new translations of non-lesson pages will also require this `original` tag. Doing this means we have a dedicated bit of metadata for finding translation sources, and leaves `redirect-from` to do its important work of ensuring good redirects without saddling it with other responsibilities.

I still have a few open tasks:

- [x] Enforce checking for `original` tags appropriately (might be a separate PR since #816 is also underway right now)
- [x] Made a little "translation map" that is mostly just useful for internal debugging, but it needs some kind of link on the site. [netlify preview](https://deploy-preview-818--ph-dev.netlify.com/translation-concordance)
- [x] render translation map as a table

Closes: #785 

Re: #651


walshbr (approved):

arojascastro (approved):
Ok, thank you for this.  I understand that you changed already the metadata of the existing translations by replacing `redirect` with `original` and that you also modified our guidelines for editors, right? Great job!

PR #153, opened 2015-11-10
Make Feedback options more visible

wcaleb:
Following up on #152 discussion, I thought it would be good to make clearer that people can offer feedback without having to create a GitHub account. 

This pull request does two main things:
- Renames "Report an Issue" page to "Feedback" page
- Makes emailing or tweeting us the first option for feedback, with GitHub options lower on the page

I also wanted to demonstrate how an internal pull request can be made by someone on the Project Team, so that if someone submits a lesson by email, we can see how discussion could still happen on a pull request if desired.

@acrymble @miriamposner @ahegel @fredgibbs @ianmilligan1 What do you think about these changes?

PR #819, opened 2018-05-02
New lesson "Getting Started w/MySQL using R"

amandavisconti:
Review ticket #153, completes all tasks moving from ph-submissions repo
to production site (still need to do Travis checks etc.)


lessons/getting-started-with-mysql-using-r.md
```diff
@@ -0,0 +1,826 @@
...
+- Jeff Blackadar
+date: 2018-05-03
+editors:
+- Amanda Visconti
+reviewers:
+- Jesse Sadler
+- Simon Appleford
+layout: lesson
+difficulty: 2
+review-ticket: 153
```
mdlincoln: Can you put the full URL in here, not just the number? (Sorry this isn't in the docs yet - still waiting on getting #816 finished!)

lessons/getting-started-with-mysql-using-r.md
```diff
@@ -0,0 +1,826 @@
...
+
+We are using the R computer language for the examples, but the techniques can be used with other languages such as Python.
+
+To do this lesson you will need a computer where you have permission to install software such as R and RStudio, if you are not running that already. In addition to programming in R, we will be installing some components of a database system called MySQL which works on Windows, Mac and Linux.
+
+Some knowledge of installing software as well as organizing data into fields is helpful for this lesson which is of medium difficulty.
+
+{% include toc.html %}
+
+# Getting Started With MySQL
```
mdlincoln: Is this h1 bit redundant? Because it's followed by an h1 Introduction?

lessons/getting-started-with-mysql-using-r.md
```diff
@@ -0,0 +1,826 @@
...
+
+Once installed, we are going to change the root password for the MySQL server.
+
+###### 3. Change the root password for the MySQL server
+
+**This section of the lesson has caused difficulty for others.  Take your time and please note that the MySQL commands end with a semicolon. Watch for them in some of the commands below.**
+
+3.1. Open a terminal window
+
+3.2. Add /usr/local/mysql/bin to the PATH per the command below. The PATH is a list of directories the computer looks in when you type a command to run a program.  In the next step below, when you run *mysql*, the PATH searches the directories it has for the *mysql* program.  PATH finds *mysql* in the directory */usr/local/mysql/bin* and runs it.  PATH just saves typing the whole path, in this case  */usr/local/mysql/bin/mysql*, to a program when you want to run it.
```
mdlincoln, line 156: Across this lesson, I've seen a few different kinds of conventions for styling text like file paths and commands that need to be entered into the console. Sometimes they're bare like this, other times _italicized_, other times **bold**.

I usually like to wrap these paths, and any kind of text that needs to go into a code file or get entered into the terminal, with `backticks so that they look like this` however this hasn't been something that we have consistently enforced across our lessons yet. 

Whichever way you go, it'd probably be good to get these strings normalized across the whole lesson text so that it's easier to understand what kind of text is "words that are code or commands that go into the console" versus "words marked bold because they're an important vocab word"

lessons/getting-started-with-mysql-using-r.md
```diff
@@ -0,0 +1,826 @@
...
+Error in .local(conn, statement, ...) : 
+  could not run statement: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'S RUDDER.',
+  '1916-05-21',
+  LEFT(RTRIM('http://newspapers.library.wales/view/4' at line 6
+```
+You can check with a SELECT statement that there is no record in the table with a story title of THE LOST LUSITANIA'S RUDDER. 
+
+Single apostrophes are part of SQL syntax and they indicate a text value.  If they are in the wrong place, it causes an error.  We have to handle cases where we have data with apostrophes.  SQL accepts two apostrophes in an insert statement to represent an apostrophe in data, as per below:
+```
+''
```
mdlincoln: Seems overkill to put this as a fenced code block - I might suggest just in-lining it as `''`

mdlincoln (changes requested):
The only thing here that _needs_ to change is getting the full URL for the review ticket. But I did have some questions about formatting the code - totally up to @amandavisconti about how you want to proceed with that!

lessons/getting-started-with-mysql-using-r.md
amandavisconti: Got it. Thank you!

lessons/getting-started-with-mysql-using-r.md
amandavisconti: Fixed!

lessons/getting-started-with-mysql-using-r.md
amandavisconti: Fixed.

mdlincoln (approved):

mdlincoln:
All looks good to me @amandavisconti - merge whenever you are ready.

Issue #825, opened 2018-05-07
Request for RSS


acrymble:
We've had a request via twitter for an RSS feed for new lessons.
<img width="305" alt="screen shot 2018-05-07 at 09 00 22" src="https://user-images.githubusercontent.com/646647/39691238-ce41169c-51d4-11e8-9fc2-c170f46a2f07.png">

acrymble:
We have this for blog posts (https://programminghistorian.org/feed.xml) - though I don't know that it's advertised anywhere.

mdlincoln:
We should definitely make this - but I'd love it if someone else took a crack at it :) I'll be glad to review a PR though.

walshbr:
I can take a look as per #828 - I'll play around and report back.

walshbr:
It looks like this actually did used to be part of the behavior of our RSS feed, but [the filtering line](https://github.com/programminghistorian/jekyll/blob/d032ccf998a0d987247de1c70bf8b109489cccd3/feed.xml#L28) is searching for lessons by whether or not they have the default template. Looks like we moved lessons to a different template without updating the feed, so we were missing the lessons and grabbing only a few junk pages. The sorting is also a bit wonky, as I think Jekyll sorts by filename and not date by default. So working on getting the most recent lessons and posts.

walshbr:
Refactored the sorting a bit on the PR based on the fact that we now have translations and will anticipate having more in the future. Will comment further on the PR but once that is merged this can be closed I think.

walshbr:
Updated in #828 - our feed has the most recent posts, original language lessons, and translations now. Though it might be worth thinking about how we could advertise it better on the site, that seems like a separate issue. So closing this request for now as it has been fulfilled.

acrymble:
@walshbr can you let the person who requested it know? Maybe good to advertise it generally on Twitter?

walshbr:
Yep can do @acrymble.

PR #828, opened 2018-05-08
Incorporate Lessons into RSS Feed

walshbr:
DO NOT MERGE YET

Testing the feed using netlify as per #825

### DIFF
--- a/feed.xml
+++ b/feed.xml
@@ -38,5 +38,35 @@
 				<guid isPermaLink="true">{{ site.url }}{{ page.url }}</guid>
 			</item>
         {% endfor %}
+
+				{% assign pages = site.pages | where:'layout','lesson'| where: 'lang', 'en' | sort:'date' | reverse %}
+				{% for page in pages limit:10 %}
+			<item>
+				<title>{{ page.title }}</title>
+				{% if page.abstract %}
+					<description>{{ page.abstract | xml_escape }}</description>
+				{% else %}
+								<description>{{ page.excerpt | xml_escape }}</description>
+				{% endif %}
+				<pubDate>{{ page.date | date: "%a, %d %b %Y %H:%M:%S %z" }}</pubDate>
+				<link>{{ site.url }}{{ page.url }}</link>
+				<guid isPermaLink="true">{{ site.url }}{{ page.url }}</guid>
+			</item>
+				{% endfor %}
+
+				{% assign pages = site.pages | where:'layout','lesson'|  where:'lang','es' | sort:'translation_date' | reverse %}
+				{% for page in pages limit:10 %}
+			<item>
+				<title>{{ page.title }}</title>
+				{% if page.abstract %}
+					<description>{{ page.abstract | xml_escape }}</description>
+				{% else %}
+								<description>{{ page.excerpt | xml_escape }}</description>
+				{% endif %}
+				<pubDate>{{ page.translation_date | date: "%a, %d %b %Y %H:%M:%S %z" }}</pubDate>
+				<link>{{ site.url }}{{ page.url }}</link>
+				<guid isPermaLink="true">{{ site.url }}{{ page.url }}</guid>
+			</item>
+				{% endfor %}
 	</channel>
 </rss>
--- a/feed.xml
+++ b/feed.xml
@@ -5,7 +5,7 @@
 <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:dc="http://purl.org/dc/elements/1.1/">
 	<channel>
 		<title>{{ site.name | xml_escape }}</title>
-		<description>{% if site.description %}{{ site.description | xml_escape }}{% endif %}</description>		
+		<description>{% if site.description %}{{ site.description | xml_escape }}{% endif %}</description>
 		<link>{{ site.url }}</link>
 		<atom:link href="{{ site.url }}/feed.xml" rel="self" type="application/rss+xml" />
 		{% for post in site.posts limit:10 %}
@@ -13,7 +13,7 @@
 				<title>{{ post.title | xml_escape }}</title>
 				{% if post.author.name %}
 					<dc:creator>{{ post.author.name | xml_escape }}</dc:creator>
-				{% endif %}        
+				{% endif %}
 				{% if post.excerpt %}
 					<description>{{ post.excerpt | xml_escape }}</description>
 				{% else %}
@@ -25,7 +25,7 @@
 			</item>
 		{% endfor %}
         {% for page in site.pages %}
-          {% if page.layout == 'default' %}
+          {% if page.layout == 'lesson' %}
 			<item>
 				<title>{{ page.title }}</title>
 				{% if page.abstract %}
--- a/feed.xml
+++ b/feed.xml
@@ -24,7 +24,8 @@
 				<guid isPermaLink="true">{{ site.url }}{{ post.url }}</guid>
 			</item>
 		{% endfor %}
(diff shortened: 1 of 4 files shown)


walshbr:
As of right now, the RSS feed in this PR grabs the most recent 10 blog posts, the 10 most recent English lessons, and the 10 most recent Spanish lessons. It could be a bit more DRY, but it wouldn't be difficult to add more translations in the future with this refactoring. Is that acceptable behavior for this @acrymble and @mdlincoln? I am not actually too familiar with RSS feeds, so I was reading around to try to see whether they tend to return everything under the sun on the site or rather just the most recent items (seemed to be the latter). Should be ready to merge as long as we're OK with those parameters and would close #825.

Also note that I found two Spanish translations that had the translation dates in an inverted format - Year-Day-Month instead of Year-Month-Day. This PR takes care of those fixes, which are necessary to get the feed to build correctly. Tagging @arojascastro just so that he can be aware that I touched some es files!

arojascastro (approved):
I am sorry. I thank you for correcting those dates.

walshbr:
No worries at all @arojascastro! Just wanted someone on the Spanish team aware I had mucked about in your files.

feed.xml
```diff
@@ -5,15 +5,15 @@
...
+		<description>{% if site.description %}{{ site.description | xml_escape }}{% endif %}</description>
 		<link>{{ site.url }}</link>
 		<atom:link href="{{ site.url }}/feed.xml" rel="self" type="application/rss+xml" />
 		{% for post in site.posts limit:10 %}
 			<item>
 				<title>{{ post.title | xml_escape }}</title>
 				{% if post.author.name %}
 					<dc:creator>{{ post.author.name | xml_escape }}</dc:creator>
-				{% endif %}        
+				{% endif %}
```
mdlincoln, line 16: Don't worry about it for this PR, but you should set your text editor to not try and automatically trim whitespace from lines.

feed.xml
```diff
@@ -37,7 +37,36 @@
...
+				{% else %}
+								<description>{{ page.excerpt | xml_escape }}</description>
+				{% endif %}
+				<pubDate>{{ page.date | date: "%a, %d %b %Y %H:%M:%S %z" }}</pubDate>
+				<link>{{ site.url }}{{ page.url }}</link>
+				<guid isPermaLink="true">{{ site.url }}{{ page.url }}</guid>
+			</item>
+				{% endfor %}
+
+				{% assign pages = site.pages | where:'layout','lesson'|  where:'lang','es' | sort:'translation_date' | reverse %}
```
mdlincoln: I don't think it makes sense to loop the Spanish lessons distinctly from the English lessons _if they're both ending up in the same feed_.

Either we should stream all the lessons and posts together (so finding all lessons without filtering for `lang == 'en'`) or we should create separate RSS feeds for lessons by language (and separate the blog posts, too). If the latter, we should make sure to use https://www.w3schools.com/XML/rss_tag_language.asp

I'm a fan of [KISS](https://en.wikipedia.org/wiki/KISS_principle) as we don't publish with ferocious frequency, and given how few users subscribe to our RSS, I don't think we have the demand for separating them into different streams right now.

mdlincoln (changes requested):
So one of those comments is really more about the bigger question of whether we have separate RSS streams. I vote no, just keep one RSS feed, but make sure lessons of all languages PLUS posts come through.

We can use [RSS category tags](https://www.w3schools.com/XML/rss_tag_category_channel.asp) to divvy them up - that way if any savvy users desperately objects to seeing blog posts, they can configure their RSS reader to do it.

walshbr:
One clarification - would we just want to list all the published lessons/translations in one big feed @mdlincoln? That's pretty easy and doable.

Or do we care about a smaller feed that only lists the most recently published items? I was having trouble with that specificity given the metadata. It seemed as though I couldn't find a good way to grab the most recent lessons that were English _or_ Spanish. In the case of the English lessons that means I am sorting on the "date" field. But for the Spanish lessons that date field isn't really the relevant metadata, as that field stores when the original corresponding _English_ lesson was published. Instead for Spanish I sort on the "translation_date" field, which the English lessons don't have at all. The logic started to feel too complicated for what we were trying to do, though I can think further and put something together if necessary.

mdlincoln:
Ooof, sorry, you're right, I forgot about the mismatched metadata names 😢  Hmm... if we could write our own jekyll plugin then it'd be easy to cross-walk everything, but since we're on GitHub we can't...

What happens if you sort first by date, and then by translation date? Does it kick the lessons up in the queue?

mdlincoln:
*adds yet another item to list of reasons to migrate the building of this site off of GH pages*

feed.xml
```diff
@@ -37,7 +37,36 @@
 				<link>{{ site.url }}{{ page.url }}</link>
 				<guid isPermaLink="true">{{ site.url }}{{ page.url }}</guid>
 			</item>
-          {% endif %}
         {% endfor %}
+
+				{% assign pages = site.pages | where:'layout','lesson'| where: 'lang', 'en' | sort:'date' | reverse %}
+				{% for page in pages limit:10 %}
+			<item>
```
mdlincoln: OK - if we can't get one unified for loop due to the date issue, then we can make this a bit more DRY by moving the `<item>` stuff into an `_include` file. The only thing that would need to differ between the includes would be the field to consult for the date... I'm hoping you'd be able to pass that as an [include parameter](https://jekyllrb.com/docs/includes/#passing-parameters-to-includes)?

mdlincoln: although i'm not sure if you can pass variable names as parameters -_-

walshbr:
@mdlincoln - "adds yet another item to list of reasons to migrate the building of this site off of GH pages" - I implemented this on my own blog recently, and it was actually pretty painless. Allowed me to keep things on github pages but also use custom plugins. It would change the branching strategy slightly so is not non-trivial, but I've been pretty impressed with it.

feed.xml
walshbr: I'll take a look at this more and play around! Knowing that I wasn't missing some obvious fix is helpful, and I'll try out these ideas.

mdlincoln:
... but GH pages expressly forbids custom plugins outside of an approved whitelist? What's the strategy you're thinking of?

walshbr:
Oh sorry - the link didn't come through - http://joshfrankel.me/blog/deploying-a-jekyll-blog-to-github-pages-with-custom-plugins-and-travisci/

It involves deploying all your files to a source branch, and then travis builds everything for you with your custom plugins and sends the built files to the appropriate branch to be deployed.

mdlincoln:
Oh, I see. For now, I'd say the new branching structure would be a complete non-starter for our team - we have a hard enough time as is just using the straightforward one! If we did ever want to migrate off, I'd probably prefer a solution like netlify that'll pull our code and build/host, meaning we don't have to change any of our day-to-day editorial practices.

BUT

that is a convo for another day I think. Do you think you can give the `_includes` strategy a try for this one and see how it goes?

walshbr:
Totally re: it not being worth it. 

Check out that commit? Think I was able to use page.lang to put the logic for the date rendering in the include itself. Much more DRY at least. Also realized that I was looping one time too many. So now we're grabbing 10 most recent posts, 10 most recent Eng publications, and 10 most recent translations. 

Tried sorting by date first and then translation date first, but then we have the opposite problem - we'd get only the translations and no recent Eng lessons. Open to other ideas!

mdlincoln (approved):

mdlincoln:
Looks great @walshbr  - thanks for taking point on this.

Issue #854, opened 2018-05-20
Internal links should use relative paths only


mdlincoln:
I'd like to update our guidelines to make it clear that internal links (aka pointing from one lesson to another on PH) should use relative links only. For example, instead of:

```
https://programminghistorian.org/en/lessons/creating-apis-with-python-and-flask
```
we should instead use
```
/en/lessons/creating-apis-with-python-and-flask
```

This makes any future site changes and page movements much simpler.

I was going to include this in the editor guidelines - do we think it should be in the author guide as well, or is that adding too much for authors?

drjwbaker:
Editors.

acrymble:
Can travis flag that at the pull request stage?

mdlincoln:
I'm still looking for a solution to that @acrymble. There are a handful of occurrences across the site we _do_ in fact want to have full links, such as in suggested citations, markdown examples, etc. If I can figure out a way to exclude errors from those examples, I will certainly add it.

Either way, we'll still want to make sure that the editor guidelines are updated so they would know why Travis is throwing errors for it.

PR #882, opened 2018-06-02
add new translation on powershell

arojascastro:
Please approve this addition. 

Links may be broken until we fix issue #854


vgayolrs (approved):

arojascastro:
Should I merge this? I think the only problem are those external links. Or should I wait until we merge issue #854

vgayolrs:
@arojascastro I'm a little out of date to answer your question...

mdlincoln (changes requested):
After merging that old PR, there's still one broken link on this new page (I think you just need to add a forward slash before `es` so it's `/es...` instead, and don't forget to include the `lecciones` segment of the path

```
- ./_site/es/lecciones/introduccion-a-powershell.html
  *  internally linking to es/introduccion-e-instalacion, which does not exist (line 873)
```

arojascastro:
I fixed that link but still is not validated.

mdlincoln (approved):

PR #859, opened 2018-05-21
Check absolute links

mdlincoln:
Closes #854 

This will check lessons for any absolute links to internal pages. 

This also includes a small edit to the text of the editor guidelines explaining that all internal links should start with `/` rather than `https://programminghistorian.org`. This change requires translation: <https://github.com/programminghistorian/jekyll/pull/859/files#diff-eb6d0485456adeb8ec93869c5023bb78>

PR #883, opened 2018-06-02
add new translation on linked data

arojascastro:
Adding new translation - please approve. Links may be broken? Thanks.


mdlincoln (changes requested):
After merging that old PR, there's still one broken link reported in this new translation:

```
- ./_site/es/lecciones/introduccion-datos-abiertos-enlazados.html
  *  External link https://es.wikipedia.org/wiki/Serializaci%C3%B3nn failed: 404 No error
```

arojascastro:
I fixed that link but still the validation is blocked.

mdlincoln:
@arojascastro yeah, at first glance it does look like one of the checks is failing - but there is a subtle difference between the `/push` Travis test, which is failing, and the `/pr` Travis test, which is passing. The `push` test ONLY looks at _the content of this branch_ - which still includes those bad 404 links that had been caused by other pages. The `pr` test, however, looks at the content of the site that would happen if this pull request were merged into the current branch - and that one passes, because in that test the site now includes those link fixes from when I merged #859.

So the short answer is, the `pr` test shows what will happen once this pull request is merged into the current site - and it is ✅

mdlincoln (approved):

Issue #855, opened 2018-05-20
Language-independent landing page


mdlincoln:
Now that all English pages are under the `/en` path except the current `index.html`, we still need to settle on what our lading page at <https://programminghistorian.org> should look like.

One possibility could be showing this intro paragraph in each language on the front page, maybe running as three columns across the screen.

![screenshot-2018-5-20 the programming historian programming historian](https://user-images.githubusercontent.com/4392604/40284130-07ef85da-5c3f-11e8-8188-e4e8ef75ff9d.png)

drjwbaker:
I fear columns would render terribly on mobile.
Rather than repeat three times, is there anyway to randomise which translation of the blurb visitors land on? We can then just put 'English', 'French', & 'Spanish' under the text.

acrymble:
Default to English page unless your language settings are set to EN or FR?

Default to English seems to be quite common for multilingual journals out of Canada: https://www.acadiensis.ca/
https://www.utpjournals.press/loi/cjh

Alternatively we could take this opportunity to re-think what is presented on our homepage. We've had a request to make the site look more like a traditional academic journal to help with cultures of credit. We could open that discussion given this opportunity.

drjwbaker:
Added policy tag because I think this is important. I suggest - per our new guidelines at https://github.com/programminghistorian/jekyll/wiki/Programming-Historian-Governance - a formal deadline for feedback of 1 June.

arojascastro:
I agree with Adam that we should "re-think what is presented on our homepage". 

I am happy with the English page as a default, but in that case it would be good that it is mentioned that is a multilingual project and that visitors may want to access to the Spanish or French version. 

In other words, the landing page should do two functions:

- introduce the english version to the english reader 
- help the spanish or french reader to get access to those versions

My 5 cents.

walshbr:
In that vein, what about defaulting to English, with translated text just below that also directs visitors as to how to navigate to the appropriate translation appropriate for them?

mdlincoln:
From 2018-05-31 call, we're generally agreed upon the idea that we want to have the front page show the brief english description of what readers will find in the site, and also make sure alternate language sides of the Programming Historian are prominently shown in the body of the page (aka user won't have to scroll or open a sidebar to see the links on the front page alone)

I'll mock up a possible design so that we have a concrete thing for feedback.

mdlincoln:
[...]

PR #893, opened 2018-06-06
Agnostic landing page

mdlincoln:
Rough trial of a simple landing page: https://deploy-preview-893--ph-dev.netlify.com/

N.b. that the "old" landing page would now live under the `/en` path here: https://deploy-preview-893--ph-dev.netlify.com/en

Closes #893 

To-dos

- [ ] tidy up the spacing - it feels a bit sparse right now, and I don't like the white band at the bottom before the footer starts
- [ ] Better mobile view - by adding the right tags into the CSS it should be possible to shrink down the relative font size for tablet/mobile viewers, so that all language links show up on one screen


vgayolrs (dismissed):

acrymble:
I'd say this is a bit bland at the minute. What if we used the sticker logos as entry buttons into the various versions of the site? Was there a motive behind removing the picture of the census counter lady?

I'm still not clear why we're advertising Github in the footer.

drjwbaker:
I like it but agree about spacing Matt (especially for mobile).

@acrymble I wouldn't say we are advertising GitHub. Rather, that is our link to the GitHub repo, which hosts our work..

arojascastro:
Thank you Matthew! 

I am not sure about "Translated lessons" because we may start publishing originals in Spanish soon. So maybe something like: "The Programming Historian is also available in Spanish and French" or use the icons as Adam suggested. 

On the other hand, the sections that we have already (Learn, Teach, etc.) could be placed after the section on language versions? Or you are planning to delete them?

drjwbaker:
> I am not sure about "Translated lessons" because we may start publishing originals in Spanish soon. 

Good spot @arojascastro :)

acrymble:
I'm putting forth an alternative suggestion (I appreciate your efforts @mdlincoln) for comment. I think we can go even more language neutral with this.
mockup-index

arojascastro:
Cool proposal too. Can we try the same but in horizontal? I do not like much the white around the blocks, but in any case it is good.

drjwbaker:
I like how language neutral this is @acrymble. But note that columns don't work on mobile https://github.com/programminghistorian/jekyll/issues/855#issuecomment-390567347

acrymble:
@drjwbaker the columns could be made responsive. Alternatively, a horizontal view (which will privilege English content on desktop devices).
mockup-index2

spapastamkou:
If we are to keep @mdlincoln 's proposal, how about keeping only each edition's title (The Programming Historian en español , The Programming Historian en français) with no intro phrase? Otherwise I would go for @arojascastro idea to ban "translated"

If we are to go for @acrymble 's proposal (which seems to me a considerable evolution but why not?), please consider some small changes to make in some phrases (and possible impact): 
édition française
Cours à évaluation ouverte par les pairs
Se lancer dans l'apprentissage (but confirmation from @mpuren would be useful)

spapastamkou:
@acrymble sorry, these are the phrases again (no need to change "version" by "édition")
version française
Cours à évaluation ouverte par les pairs
Se lancer dans l'apprentissage (but confirmation from @mpuren would be useful)

amsichani:
I kinda like the vertical bars @acrymble suggests - you get all available languages/versions in one glimpse in a clear layout. I don't think is necessary to add any text inside the landing page logos, mainly because are just for navigation purposes (and not so much descriptive/informative) + it might mess up things in other devices.

mpuren:
@spapastamkou I will go with:
Cours à évaluation ouverte par les pairs
Débuter l'apprentissage (des humanités numériques?)
What do you think? It seems weird to me to say something like "Se lancer dans l'apprentissage"

spapastamkou:
@acrymble please consider use of the following for the FR part, thank you very much @mpuren
Version en français
Cours à évaluation ouverte par les pairs
Débuter l'apprentissage

arojascastro:
I have been thinking about this issue, and I think the logos design in three columns or bands is very radical. I know I should argue for this big change that highlights language and puts our cultural differences in the center, but as *user* I do not like to be put in a box so easily. Imagine: I visit the Programming Historian, I do not considerer myself bilingual but I think in English as much as in Spanish and certainly my working language is English. So the first thing that I am asked to do is to select a language... I do not know, I think people are more relaxed about languages. I envision a home site that informs about the project and the available editions / versions, but that does not force me to identify myself with a language or country but to *travel* or *walk* through its contents in multiples languages in non exclusive manners. Sorry if this is too long and vague.

mdlincoln:
This is really great feedback, and I really like @acrymble's color blocking.

That said, I do see how the color blocks design (which is very handsome indeed) appears to emphasize that the different languages are strictly divided. Perhaps if we try making the different color blocks contiguous it might help?

While I think I get some of the abstract things you're saying @arojascastro, lessons in different language _are_ practically divided up into different branches of the site, so at that level, it's inevitable that we have to offer different paths when someone comes to the root page of the site. That said, the language menu in the upper right corner is now much more integrated, so if you are on a page with a translation, it will send you to the equivalent page in the other selected language instead of sending you all the way back to the language home page. Perhaps there's a way for us to style the language selection menu in the header so that it helps out with this.

acrymble:
Can we use people's browser language settings to redirect them or at least present contextual information about what this website is, privileging the language they choose on their computer? Default to English unless it's Spanish or French? Maybe that takes on board some of @arojascastro 's vision?

mdlincoln:
@acrymble It's a neat idea, but it's important that we try to stick to one thing at a time right now. As a group, we've already committed to the core requirements for this landing page. Let's first get that working alongside the new French pages, and then we will have the bandwidth to experiment more.

mdlincoln:
Also I'm closing this particular PR in favor of the color blocking method. Once we get #899 fulfilled I'll try to push up another mockup.

PR #906, opened 2018-06-11
Update reviewer-guidelines.md

drjwbaker:
Or maybe it should only be Amanda? (I note the ES translations https://programminghistorian.org/es/guia-para-revisores only lists ES editors in this role)


acrymble (changes requested):
It should probably be just Amanda for English speakers. Either way, Maria José's first name only is listed, so if you're going to put here on there too you need her whole name.

drjwbaker:
@acrymble Fixed to Amanda only. Not sure how I missed of Maria José's full name. My bad!

acrymble (approved):

PR #913, opened 2018-06-16
repair markdown links that had extra whitespace

mdlincoln:
A few of our links accidentally put spaces between `[href text] (href url)` which meant they did not render properly. This PR fixes those links.


arojascastro (approved):
Thank you! Although I still do not understand fully what was wrong here.

mdlincoln:
@arojascastro one cannot put a space between the `[...]` and the `(...)` parts of a link in markdown.

PR #914, opened 2018-06-17
Color blocks landing page

mdlincoln:
A revised attempt at #893 

This PR is just for preview purposes so we can see what the French version would look like. I'll generate another with just EN and ES when we're satisfied with it.

Preview: https://deploy-preview-914--ph-dev.netlify.com/


drjwbaker:
Do we have a preview link for this?

walshbr:
@drjwbaker - https://deploy-preview-914--ph-dev.netlify.com/

drjwbaker:
Ta. Played with it on a few different laptop screen sizes and OSs. I like it a lot. Only wrinkle: the columns need to be shorter as sometimes the 'Start Learning' et al click through link is below the bottom of the page when the page loads.

drjwbaker:
Got a lay person feedback on this design: if we don't make each colour block a linked click through, we should make the bit below the white line - the bit that separates, eg, the English ISSN and 'Start Learning!' link - a big button that fills that space with the wording like 'Start Learning!' in it.

acrymble:
Thanks for user testing @drjwbaker. I think we should do more of that. It may not be clear to a newcomer what we are from this approach. But we won't know unless we ask some people.

_data/snippets.yml
```diff
@@ -327,7 +327,7 @@ peer-reviewed-lessons:
   fr: Peer Reviewed Cours
 start-learning:
   en: Start Learning!
-  es: Comencemos aprediendo!
+  es: ¡Comencemos a aprender!
```
mdlincoln: Thanks for the fix @arojascastro !

mdlincoln:
I'll toy with image sizing and seeing about ways to make it a bit more clickable.

mdlincoln:
I'll probably be redoing this from scratch, so I'm closing this PR now.

PR #922, opened 2018-06-22
add new translation on r

arojascastro:
Attempting to add a new translation by @jenniferisasi 

Please approve after checking TRAVIS validation process.

PR #923, opened 2018-06-22
Remove redirect from metadata lesson

arojascastro:
As discussed here https://github.com/programminghistorian/jekyll/pull/922 `redirect` is not longer used, so I would like to remove it from here.

### DIFF
--- a/es/lecciones/analisis-de-corpus-con-antconc.md
+++ b/es/lecciones/analisis-de-corpus-con-antconc.md
@@ -19,7 +19,6 @@ translation-reviewer:
 review-ticket: https://github.com/programminghistorian/ph-submissions/issues/170
 layout: lesson
 original: corpus-analysis-with-antconc
-redirect_from: /es/lessons/corpus-analysis-with-antconc
 difficulty: 1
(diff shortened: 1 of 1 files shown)


mdlincoln:
😬 Sorry I left a confusing comment. We do not need `redirect_from` in _new_ translations, because my changes to the way that we point to translated lessons made that workaround obsolete. 

However, we do need to keep `redirect_from` in old translations that have already been published using a `redirect_from`, because once we publish a redirect link, we need to make sure that it keeps working. For example, If someone copied the URL `https://programminghistorian.org/es/lessons/corpus-analysis-with-antconc` from our public page already, we want to make sure that their link will still work.

For new lessons like #922, we can skip adding `redirect_from`, which means that a link like `https://programminghistorian.org/es/lessons/basic-text-processing-in-r` will never be published.

Sorry this is confusing 😢

arojascastro:
Ok,  I see! But should I say something about that in our guidelines? I understand that from now on we will not used that metadata field, right? So the guidelines should not cover it perhaps?

PR #974, opened 2018-08-10
Add unity lesson

walshbr:


walshbr:
There were a few oddities with this one. Can you take a look at the points below @mdlincoln when you get a chance?

1) my local copy of the repo kept insisting that researchpapers/Afanador_Paper_DH2018_The_Programming_Historian_en_Español.pdf was a file not tracked in the repository, even though it only appeared on the local copy at all because I pulled it from the remote. Couldn't get it stop giving me those warnings, nor could I get it to go away. Nor could I get it to actually unstage once I had accidentally stage it. I am open to suggestions about this…
2) This lesson is in a weird place in that it was a reworking of a previous lesson. I've left the old editors and reviewers in the header and added the new ones (since the content is substantially the same, it seemed that those people should get credit). I also left the abstract the same @jacobwgreene, but let me know if you want otherwise. Easy enough to update.
3) Because it is both a new lesson and an update on an old, technically it has two review tickets. But the current review ticket seems to be the most relevant, and I link to the original review ticket from within that new conversation at the top of the ticket. So I just left the one ticket linked in the YAML. Let me know if that doesn't work?

Other than those things we should be good to merge.

mdlincoln:
The file name problem is a resurfacing of that old unicode filename problem. I have forcibly changed it and pushed those changes to `gh-pages` so everyone will have them, and then updated this branch with those changes as well.

en/lessons/creating-mobile-augmented-reality-experiences-in-unity.md
```diff
@@ -0,0 +1,411 @@
...
+| **Minumum Hardware Needed**           | iOS or Android Mobile Device    | Laptop/Desktop and iOS or Android Mobile Device |
+
+
+### Installing Unity and Vuforia
+
+Since the release of Unity 2017.2, the Vuforia SDK is integrated into the Unity Editor. The [Vuforia SDK](https://developer.vuforia.com/) is a suite of assets and code snippets that work alonside Unity to make it easier to quickly develop and build AR content for a variety of platforms. If you are unable to download Unity 2017.2 or later, [consult this archived lesson for earlier versions of Unity](link to previous lesson). To download Unity and Vuforia, go to the [Unity website](https://unity3d.com/get-unity/download/archive) and download Unity 2017.2. In the Components dialog box, select Vuforia Augmented Reality Support in addition to either Android Build Support and/or iOS Build Support, depending on your target mobile device platform. Once the download completes, start Unity and follow the setup prompts. If Unity asks if you are creating a personal or professional account, choose personal account.
```
mdlincoln: the `(link to previous lesson)` needs a URL

mdlincoln (changes requested):
Just one broken link to fix it looks like. I think your other workarounds make perfect sense.

en/lessons/creating-mobile-augmented-reality-experiences-in-unity.md
walshbr: Yep just noticed that. Will fix in a sec.

mdlincoln (approved):

mdlincoln:
Looks good now :) @walshbr merge whenever you're ready

Issue #957, opened 2018-08-02
Geocoding Historical Data using QGIS - truncated data file


sharonhoward:
The AlumniCounties.csv file attached to this lesson appears to have got truncated somewhere during the publication pipeline. It ends at Merionethshire. I compared it to the version in ph-submissions/assets/ and that file looks complete (it has a further 24 rows of data).

acrymble:
Thanks @sharonhoward I will check it out when I can.

acrymble:
this has been fixed. Thanks @sharonhoward

PR #980, opened 2018-08-16
Issue #957 - truncated data files in Geocoding QGIS lesson

acrymble:
@sharonhoward pointed out that the csv data file in https://programminghistorian.org/en/lessons/geocoding-qgis was incomplete, but that the complete data file was available on the submission site.

This was obviously a mistake so I've added the missing lines of data. I've also moved the files to a dedicated folder in the assets directory as per our current practices:

/assets/geocoding-qgis

I've checked and it doesn't look like this has been started yet on the /ES branch, so this shouldn't affect them. @spapastamkou and @mpuren does this affect the /FR branch?


acrymble:
This closes #957

spapastamkou:
not in the current state of the /FR branch. 
The tutorial is in the list of the students' translations, if modifications are needed, we can do them as editors before making the text available for review (in some weeks from now).

acrymble:
Thank you @spapastamkou

mdlincoln (approved):

Issue #939, opened 2018-07-15
possible typos in R with Tabular Data lesson


acrymble:
following has been reported as potentially wrong in the R with Tabular data lesson (with thanks to Heather Bennett:

In the Matrices section, I spotted three things:

1. When explaining the apply() function, the author writes: "If you want to find the mean of each row, you would use 2 instead of 1 inside of the function.” I think this is backwards? Based on the examples given it looks like 1 is for rows and 2 is for columns.

2. In the practice segment, instruction 2 states: "use the cbind90 function…” I think this is meant to be cbind()

3. Ditto in instruction 3: “Calculate the mean of each column for the above matrix using the apply90 function”. I’m assuming “apply90” should be “apply()”

arojascastro:
We are reviewing the translation of that lesson here: https://github.com/programminghistorian/ph-submissions/issues/164

Can you @jenniferisasi have a look at this? Gracias.

acrymble:
Thanks @arojascastro that would be amazing to get another opinion.

jenniferisasi:
Hi! 

- 2&3 Yes, that 90 should be a () ; cbind() and apply()
I have cbind() and apply() on my translation, I think I automatically put the () after the functions. 

- 1 I'm not an expert and I assumed the tutorial was right, but, yes, with ?apply in R it indeed says: 
> MARGIN a vector giving the subscripts which the function will be applied over. E.g., for a matrix 1 indicates rows, 2 indicates columns, c(1, 2) indicates rows and columns.

acrymble:
@jenniferisasi thank you for confirming 2 & 3 need fixing. I've fixed those and will submit a pull request.

For 1, Heather is correct and the text needs to be changed. In that paragraph the author has columns (which go down) and rows (which go across) backwards. So where it says column it hsould say row, and vise versa. The updated English text is:

> The above example shows the apply function used on the `Crime` matrix to calculate the mean of each row, and so the average number of combined theft and violent theft crimes that were committed in each decade. If you want to find the mean of each column, you would use `2` instead of `1` inside of the function:

Please make sure the translation is updated before it goes live.

PR #981, opened 2018-08-16
typo fixes r-basics with tabular data as per #939

acrymble:
As noted by Heather Bennett, there were 3 small errors in https://programminghistorian.org/en/lessons/r-basics-with-tabular-data

This pull request corrects them. I've also asked @jenniferisasi to make sure they are fixed in the Spanish translation, which is under review.


arojascastro (approved):
The translation is correct

PR #1005, opened 2018-09-03
Matthew Lincoln Revised bio

mdlincoln:
I've updated my bio - the translation will need to be reviewed. Thank you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ _site/*
 .jekyll-metadata
 .DS_Store
 tmp/
+
+vendor/
+.bundle/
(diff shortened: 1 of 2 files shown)


acrymble:
This is the line that needs translating:

- Matthew Lincoln is the digital humanities developer at Carnegie Mellon University, and an art historian of early modern Europe.

I'm not a Spanish speaker, but I think we want this:

- Matthew Lincoln es desarrollador humanidades digitales en el Carnegie Mellon University e historiador del arte de la Europa moderna.

jenniferisasi:
If my changes past the test, we should be ready to merge.

jenniferisasi (approved):

PR #1008, opened 2018-09-09
New page of lessons we'd like to see

### LINKED ISSUE #995: Page of lessons we'd like to see
Suggestion coming out of our community survey #576:

- maybe list lessons you would like filled so that less confident authors might feel inclined to write one or collaborate with someone else

We have this on our blog from time to time, but I think this is a good suggestion. To make it easier to keep in sync, I suggest 1 page for all 3 languages, listed by "Requests from X Team".

acrymble:
As per issue #995 and a request from the user survey.

This is in English only because no Spanish suggestions came forth. A Spanish page can be added in future if they would like one.

I tried to add this to the /contribute menu so that needs checking.

Closes #995


walshbr (changes requested):
I'd add two line breaks after the title of number 7 to conform with the formatting for the other suggestions. Other than that it looks good!

acrymble:
@walshbr thanks, I've fixed that.

mdlincoln (commented):
The one problem with putting this as a page within the menu bar is that the menu bar is designed with the assumption that all links are visible on all sites. Right now, if I look at the menu in https://deploy-preview-1008--ph-dev.netlify.com/es/ it looks like so:

screenshot_2018-09-09 the programming historian en espanol programming historian

This is pretty deeply coded into the design of the site layouts, so I wouldn't want to put in a workaround to remove that blank bit of menu box on the ES and FR menus _just_ for this page.

The alternative is to have links to the EN lesson request page display on all languages, just like the "Blog" link shows up for all pages but only goes to the one blog page. So, for example, there'd be a link for "solicitudes de lecciones" on the ES site, but it would end up at the EN page for requests... at least for now, until we do end up with ES lesson requests.

What do you think?

mdlincoln:
FYI if you do want to go with the "Blog"-like option I described, you'd just need to add in translated titles to the `snippets.yml` file, but make the links all point to `/en/lesson-requests`

acrymble:
@mdlincoln I think having them all arrive at the single page is fine. How do I do that?

_data/snippets.yml
```diff
@@ -54,6 +54,10 @@ menu-contribute-overview:
   es:
     title: Contribuciones
     link: /es/contribuciones
+menu-lesson-requests:
+  en:
```
mdlincoln, line 58: @acrymble to add links all pointing to the same page, repeat this yaml block for `es` and `fr`, just copying the same `link` value each time. It would probably be good to translate the `title` values, though.

acrymble:
@walshbr I've made your changes. Can you accept them?
@mdlincoln I've also made your change to the snippets file as requested.

walshbr:
@acrymble I made one change - the way that markdown parsing numbers lists was making each new item in the list restart the list at number one. I fixed it by following this - https://stackoverflow.com/questions/18088955/markdown-continue-numbered-list. Should be good to go now as far as I am concerned. Good to merge if good with @mdlincoln

walshbr (approved):

PR #1026, opened 2018-09-22
Update guia-para-autores.md

mariajoafana:
I fixed a mistake on my email address


acrymble:
I probably made that mistake in the other files too.

acrymble (approved):
Looks like the only wrong email address. Good catch!

PR #1029, opened 2018-09-24
Links to images should be https

mdlincoln:
Closes a minor technical problem where I was getting page build warnings because some of our https pages were embedding images that used http links instead.

### DIFF
--- a/_plugins/validate_yaml.rb
+++ b/_plugins/validate_yaml.rb
@@ -28,7 +28,7 @@ module MyModule
       valid_difficulties = [1, 2, 3]
 
       # Fields required on ALL lessons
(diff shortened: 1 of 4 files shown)


mdlincoln:
Please don't merge yet... I'm testing out a fix that should validate this on our side rather than rely on github/netlify to complain...

mdlincoln:
OK, there's now a new step in the validation script that checks for insecure image content. There were virtually no examples of this because most of our images are self-hosted, but on occasion we do hotlink out to e.g. wikipedia, so this will rarely if ever be flagged.

walshbr (approved):

PR #1035, opened 2018-10-03
Add Treasurer and Sponsorship roles

drjwbaker:
Edits required to add Treasurer and Sponsorship roles to the site. Needs ES translation and to be added into FR translation work.

PR #1044, opened 2018-10-07
Translate treasurer and sponsorship coord roles

vgayolrs:
Hi @acrymble @drjwbaker @jenniferisasi @spapastamkou !

As I said here #1035 , the terms to refer to the person in charge of the Treasury are "tesorero" for men and "tesorera" for women. In such a way is that I chose Treasury (the function). In the other hand, Coordinator of sponsorship is "coordinador" (male) and "coordinadora" (female), and I chose "Coordinación de patrocinio". If the person is already designated, I suggest modifying the term according to gender. The mechanism will be the same in French as the language differentiates genres.

Happy to hear any thought!

### DIFF
--- a/_data/teamroles.yml
+++ b/_data/teamroles.yml
@@ -78,7 +78,7 @@
 - type: sponsorship
   label:
     en: Sponsorship Coordinator
-    es: Cordinación de patrocinio
+    es: Coordinación de patrocinio
   definition:
     en: Responsible for managing relationships with our sponsors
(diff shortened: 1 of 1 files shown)


drjwbaker (approved):

drjwbaker:
@vgayolrs Sounds reasonable. Thanks to bringing to our attention the gender issue. Given the way grammatical gender work in ES and FR compared with EN, we might need to be more attentive to this in our web/data design in the future.

PR #1055, opened 2018-10-17
Markdownify abstracts

### LINKED ISSUE #1054: markdown not rendered properly in lesson abstracts on lesson page
screen shot 2018-10-16 at 8 54 17 pm

I need to sneak a call to `markdownify` in the right template...

mdlincoln:
Closes #1054 

Almost none of the lessons use markdown in their abstracts... but when they do, this ensures that it displays ok.


arojascastro:
Very good! I encountered that problem recently. So thanks.

arojascastro (approved):

PR #1092, opened 2018-11-25
Typo fixes for jq lesson

### LINKED ISSUE #1091: Improvements for "Reshaping JSON with jq"
- The full title of the lesson: `Reshaping JSON with jq`
- The system you are using: `Mac`

# Core jq filters
- Since there are references to the filename `jq_rkm.json` elsewhere, and the target of the link is to a file named `jq_rkm.json`, it might be clearer to have the anchor text be `jq_rkm.json` also.

# The Pipe: `|`
- In first paragraph, typo: `.artOjbects[]` should be `.artObjects[]`
- `every value at the key id within the artObjects array`: might be clearer to say `artObjects array ` _`elements`_ here

# Create new JSON: [] and {}
- Might be clearer to demonstrate creating a new object with _different_ keys, e.g. ` {painting_id: .id, painting_title: .title}`. Otherwise, the existing example can be simplified to just specifying the field names directly, like this: `.artObjects[] | {id, title}`

# JSON vs JSON Lines
- There is a website describing this defacto format: http://jsonlines.org/

# One row per tweet
- In the final example demonstrating how to separate hashtags with semicolons, it mentions "There are ways to get the same results using an even shorter query". Perhaps an example, such as this, could be given, for further study:

```
[.id, ([.entities.hashtags[].text] | join(";"))] | @csv
```

# Extracting user data
- Typo: `numers` should be `numbers
- Typo: `deliminters` should be `delimiters`

mdlincoln:
Closes #1091

n.b. this lesson has not been translated


acrymble (approved):
inspired contribution

PR #1113, opened 2018-12-03
Updating bio and adding an avatar for @fdlaramee; Closes Issue 1112

fdlaramee:
As requested in onboarding ticket 1096, I have:

* added a black and white avatar to the /avatars directory
* upgraded my bio in the /data/ph_authors.yml file


acrymble:
It's showing an error in the travis checks. I wonder if this is because there's an accent in the filename? It says the file doesn't exist, but that's because you're uploading it...

fdlaramee:
@acrymble Yes, there is a missing accent somewhere. I am trying to fix the issue right now.

fdlaramee:
@acrymble The checks have passed now. Seems like setting "team: true" in my bio broke the process (the system probably assumes that the avatar is in the live repository for team members?) so I put it back to false and it solved the problem.

spapastamkou:
I am not very sure what to do and I would appreciate a little help here (@acrymble ?). It has to be "team: true" if FD is to appear as member of the team, right? Am I to change it and then approve?

acrymble:
I don't know. @mdlincoln is the best person to ask.

walshbr:
The way we handle diacritical marks in names is a little counter-intuitive - we strip them out of the filename for the avatar image itself but preserve them in the authors file. Then we have a little bit of logic in the contact-info include that converts the diacritical version to the version that appears in the avatar filename. That commit should fix things up.

mdlincoln:
oh jeez, [I (circa 2017) am apparently to blame for that insane chain of formatting](https://github.com/programminghistorian/jekyll/commit/cc46c6dfa#diff-0667089f0a33064fd7e6205a942f0e7d) :(

We live and learn, though. I'm now aware of a much better solution for this: [the `slugify` Liquid filter](https://jekyllrb.com/docs/liquid/filters/), whose `"latin"` option does this job for us cleanly.

mdlincoln:
... and although most of the webservers don't care about the capitalizations of the avatar files, our htmlproofer does... i'll poke around a bit more but for the time being we'll take @walshbr's patch and get your bio published!

mdlincoln (approved):

PR #1139, opened 2018-12-21
Standardize gallery image sizes

### LINKED ISSUE #1138: Inconsistent Gallery Image Sizes
Our gallery images at present are inconsistent. The editorial guidelines state things should be a square (about 98% of them were) but they also state that they should be scaled down to 200x200. A lot were squares larger than that, which was causing inconsistent design for the lesson pages themselves.

walshbr:
Closes #1138.

I think I checked everything to make sure this didn't cause any weirdness or grab any images that we didn't want resized (banner/splash images).


mdlincoln (commented):
Thanks for fixing this, and sorry I'm only just looking at it now @walshbr!

On top of fixing up the images, I wonder if there's some tweak we can make to the CSS or to the image classes so that the avatars are always consistently-sized? Perhaps creating a class `lesson-avatar` and giving it a setting of `max-width: 200px;`? What do you think?

walshbr:
@mdlincoln - I'll take a look! Where are they inconsistently sized? On the project team page? And do you mean consistently sized as the window scales down? Or relative to itself just within one window size?

mdlincoln:
No, all the images you uploaded are ok. What I meant was configuring the CSS so that if anyone happens to upload, say, a 250px wide image _in the future_, then the display on the lesson page, as well as the lessons list page, works out ok.

walshbr:
Gotcha. How about this then? Added max-width to a lesson-avatar class on the lesson images only on the individual lesson pages.

mdlincoln:
silly me, the width of that image is actually controlled by a wrapper `div` that already has a class. I put the `max-width` constraint there.

mdlincoln (approved):

PR #1164, opened 2019-02-10
Language button group selector

### LINKED ISSUE #1104: Enhanced language switcher on menu bar
I think it'd clarify what hitting the en/es/fr links in the menu bar would do if we had a button group that displayed all three languages and highlighted the active one, a la https://getbootstrap.com/docs/4.1/components/buttons/#checkbox-and-radio-buttons

mdlincoln:
This PR creates a button language selector in the nav bar that always shows the available languages.

Closes #1104

### DIFF
--- a/_data/snippets.yml
+++ b/_data/snippets.yml
@@ -144,13 +144,13 @@ menu-blog:
     link: /blog
 menu-lang:
   es:
-    title: '[es]'
+    title: es
     link: /es
   en:
-    title: '[en]'
+    title: en
     link: /en
   fr:
-    title: '[fr]'
+    title: fr
     link: /fr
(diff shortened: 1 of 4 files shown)


mdlincoln:
One question I'm not entirely sure about is whether I've set up the `active/inactive` and `aria-*` attributes properly so that the links are understandable by a screen-reader. I tried to parse the Bootstrap docs as best I could but I'd love your thoughts @ZoeLeBlanc

ZoeLeBlanc (approved):
Looks good to me! TBH I don't know much about accessibility but I read up in the MDN and Bootstrap docs and your changes look solid 👍

PR #1175, opened 2019-02-16
Create lecons file

spapastamkou:
File for the FR tutorials once they are ready after review.


mdlincoln:
Thanks for opening this - I added the folder in a slightly different way (adding an empty folder in git is not very intuitive) so you should now see it on the `gh-pages` branch.

PR #1176, opened 2019-02-18
add keepfile to lecons

mdlincoln:
re #1175


spapastamkou:
Thank you, @mdlincoln !

PR #1197, opened 2019-03-01
Metadata spacing

### LINKED ISSUE #1106: even out spacing of lesson metadata
This uses bootstrap `col`s right now but should probably use `flexbox` instead for better justification.

mdlincoln:
This closes #1106 by using bootstrap's flexbox utilities instead of relying on a complicated series of if/else tags to place lesson metadata based on whether the lesson was an original, or a translation (and thus had a lot of additional contributor names to attach to it)

The only remaining bit of if/else logic comes for original lessons, giving a more defined space to the editor and reviewer metadata.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -5,11 +5,11 @@
 <footer role="contentinfo" style="background-color: {{site.data.snippets.background_color[page.lang] }}">
 
 
-<div class="d-flex justify-content-center footer-head">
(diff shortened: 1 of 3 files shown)


ZoeLeBlanc (approved):
Weird flex but ok 😂 

jk couldn't resist a flexbox pun. Looks great 👍

PR #1212, opened 2019-03-09
All pages plugin check

### LINKED ISSUE #1203: build plugin should check all pages for absolute links, not just lessons
Right now there are routines that run for lessons and others that run for blog posts - they check yaml fields specific to those page types. Should make one that runs over all pages to check

- absolute links
- insecure image links

mdlincoln:
Closes #1203 

This will check all pages - not just lessons or blog posts - for absolute links and insecure image content.


ZoeLeBlanc (approved):
Looks good to me. Really like the changes to the `validate_yaml` script 👍

PR #1214, opened 2019-03-10
Fix invalid Turtle examples

dbs:
Each prefix directive in Turtle requires an ending period.

If the subject is the same, but the predicate and subject are different, then
you have to use a semicolon for a short form statement.

Signed-off-by: Dan Scott <dscott@laurentian.ca>


arojascastro:
Thank you! We will review this suggestion in the following days. 

For editors: if the change is accepted, apply in Spanish translation as well.

mdlincoln (dismissed):

PR #1252, opened 2019-04-05
sort by translation date if it exists, otherwise use pub date

### LINKED ISSUE #1251: Lesson Index translations+original
Just got to our main page to check our new and first original lesson and found a technical issue:
- If organized by the default new to old, it shows the original at the end (!!)
- If switched from old to new, it shows the original then translations oldest to newest. 

As we move forward with translations from Spanish to English, and potentially future original lessons in French, the 3 sites need a way to display both originals and translations in order (I don't think separating them will be good, but it'd depend on the technical solution) 

I'm afraid I need to ping @mdlincoln and @ZoeLeBlanc, and ask, is there a way to "merge" the yaml fields for translation and publication? 

- We also now need to change the words in this tab:
Screen Shot 2019-04-04 at 2 47 10 PM

mdlincoln:
Closes #1251

This adjusts the code that creates the "sort date" when rendering a lesson page. IF a lesson has a translated date, that's the sorting date. If it is an original lesson without a translation date, then the sort date will be the publication date.

@programminghistorian/spanish-team needs to update the `sort-by-date` entry in `_snippets.yml`.

### DIFF
--- a/_data/snippets.yml
+++ b/_data/snippets.yml
@@ -217,7 +217,7 @@ sort-by-date:
   en:
     sort by publication date
   es:
-    ordenar por fecha de traducción
+    ordenar por fecha de publicación
   fr:
     trier par date de publication
 sort-by-difficulty:
--- a/_includes/lesson_describe.html
+++ b/_includes/lesson_describe.html
@@ -26,7 +26,7 @@ lesson-slug.html computes the correct lesson slug, and then this include determi
   <p class="abstract">{{ lesson.abstract | markdownify }}</p>
 
   {% comment %}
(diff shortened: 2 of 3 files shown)


mdlincoln:
While I'm at it, I've also added a check that `translation_date` is greater than `date`

jenniferisasi (dismissed):
Great! And thanks for the comments you enter on the code itself, is very helpful!

jenniferisasi (approved):

jenniferisasi:
Approved @mdlincoln's changes to sort publications and changed snippet to reflect original publications in spanish.

PR #1262, opened 2019-04-09
only load MathJax on pages that need it

mdlincoln:
This adds logic to our html headers so that MathJax only gets loaded on pages that set `mathjax: true` in their header files. Loading MJ unnecessarily definitely slows down page load, so this should help keep PH relatively lightweight for our visitors.

(I believe the remaining JS libraries are all used for the dropdown menus as well as for the "last updated" query that pings the GitHub API, which appears on every single page)


ZoeLeBlanc (approved):
looks 👍 
nicely done 🎉

PR #1268, opened 2019-04-12
Update 2019-04-06-PH-contributors.md

drjwbaker:
Correct final map caption (it currently is the same as map 2, which is incorrect)


acrymble (approved):
thanks

PR #1269, opened 2019-04-12
Update pic

arojascastro:
Hi, my pic was too old. This one is also old, but I look better. I used the same name, so it shouldn't break anything.


acrymble:
The image needs to be black and white.

arojascastro:
oh yes, right

mdlincoln (approved):

mdlincoln:
✨

Issue #1181, opened 2019-02-20
possible bug in "using geospatial data"


mdlincoln:
[reported from twitter](https://twitter.com/Nate_Housley/status/1097272622834798598):

>There seems to be a directory missing in the data provided for "Using Geospatial Data to Inform Historical Research in R." The tutorial reads: Census_Data <- read.csv("./data/County1990_Data/nhgis0027_ts_nominal_county.csv" but there is no County1990_Data folder

I can look into it this weekend unless someone has a chance to take a look sooner.

mdlincoln:
OK, I've checked through the lesson and the uploaded data have quite a few problems.

Firstly, our reporter is correct - there is no `Country1900_Data/nhgist0027_ts_nominal_country.csv` file.

Moreover, the `County1900` directory data is incomplete - it has data, but no spatial information, such that when I run the very first import command of the lesson, I get the following warning:

```r
County_Aggregate_Data <- st_read("./data/County1990/")
#> Reading layer `US_county_1990' from data source `/Users/mlincoln/Development/geospatial/data/County1990' using driver `ESRI Shapefile'
#> Warning message:
#> no simple feature geometries present: returning a data.frame or tbl_df 
```

Because of this, none of the rest of the lesson will work. FWIW I also tried opening this shapefile in QGIS (which also ought to be able to read it) and it says that it's not a valid shapefile, so whatever data have been uploaded are corrupt or incomplete.

@JMParr you were the editor for this lesson, can you please check if you have the complete data for it somewhere?

JMParr:
Hi, @mdlincoln - I've looked through the files, including old clones from Github. Unfortunately, it looks like I do not have complete data. I'm not sure what happened. Correcting it will necessitate contacting the author.

mdlincoln:
@JMParr ok, please contact them. Without it the lesson is unusable.

acrymble:
I'm going to put a warning message on this until this can be resolved. We don't want people thinking this lesson is useable.

acrymble:
@ladew222 you were the author of this tutorial. It looks like there has been an issue getting all of the data files in place. Can you help us rectify this so we can re-publicize the lesson?

mdlincoln:
Thanks for adding that warning @acrymble, I think that's a reasonable interim action to take. @JMParr have you emailed the author yet? 

[...]

PR #1312, opened 2019-05-23
Issue 1181 - missing data from Geospatial Analysis lesson

acrymble:
This pull request should add the missing files, as well as includes 2 updates to code lines submitted by the author. This was a confusing situation, so the pull request needs someone to actually try the lesson again to make sure it all works. I am submitting a pull request so that we can preview it easily.

re: #1181


mdlincoln:
Sorry for the delay in getting to this - I will be able to take a look tomorrow.

mdlincoln (changes requested):
Thank you for your patience.

Regretfully, the new data are not acceptable in their current form. 

The new shapefile geometries are invalid, and thus the data are unplottable.

```
> library(sf)
> library(tmap)
> County_Aggregate_Data <- st_read("data/County1990/")
Reading layer `US_county_1990' from data source `/Users/mlincoln/Development/bridges/konigsbergr/data/County1990' using driver `ESRI Shapefile'
replacing null geometries with empty geometries
Simple feature collection with 3141 features and 20 fields (with 3140 geometries empty)
geometry type:  POLYGON
dimension:      XY
bbox:           xmin: 1584238 ymin: 409629.1 xmax: 1666567 ymax: 467600.4
epsg (SRID):    NA
proj4string:    +proj=aea +lat_1=29.5 +lat_2=45.5 +lat_0=37.5 +lon_0=-96 +x_0=0 +y_0=0 +datum=NAD83 +units=m +no_defs
There were 50 or more warnings (use warnings() to see the first 50)

plot(County_Aggregate_Data$geometry)

Error in plot_sf(x, ...) : 
  NA value(s) in bounding box. Trying to plot empty geometries?
```

Note the warnings when reading the shapefile. It is possible that this may have occurred when trying to simplify the geometry of the shapefile. Rather than simplifying the geometry for the entire US county file, I recommend the author provide a shapefile that is _already_ subset to just North and South Carolina, without trying to do any geometry simplification, and that the author rewrite the intro to match the newly provided data. I've tested this and it's just 5mb when zipped.

[carolina1990.zip](https://github.com/programminghistorian/jekyll/files/3232972/carolina1990.zip)

Further down in the lesson, the command

```r
geocoded_addresses <- read.csv("data/GeocodedAddresses.csv", as.is=TRUE)
```

fails because there is no file `GeocodedAddress.csv` in the supplied zip file. I'm going to stop this review here since entire data files are still missing.

mdlincoln:
the geocoded csv seems to have been uploaded as a separate file instead of put in the archive. I'm still going to wait to continue the review until we can get an acceptable shapefile.

acrymble:
I've sent this back to the author.

Issue #1309, opened 2019-05-22
Update /research with new publications


acrymble:
I believe colleagues @amsichani @drjwbaker @mariajoafana @walshbr ? recently published something. Can I ask them to make sure this ends up on /research in all 3 languages?

Anyone else with research or outputs please also update this page so we have a full record of activity.

acrymble:
This has been added.

PR #1314, opened 2019-05-24
Add Sichani et al

drjwbaker:
Per https://github.com/programminghistorian/jekyll/issues/1309 add EN/FR/ES text for Sichani et al.


acrymble (approved):
Thanks James

PR #1334, opened 2019-06-01
add aria roles to navigation dropdowns

### LINKED ISSUE #1333: Add ARIA roles to menubar dropdowns

mdlincoln:
I've tried to add aria `role` attributes to all of the nav bar elements so it's clear which clickable things open menus vs. which are links that go to new pages. 

closes #1333


walshbr (approved):
Looks great!

PR #1356, opened 2019-06-27
Remove a tags surrounding images

### LINKED ISSUE #1325: Do not target images with links
re: #1315 

Right now, our figures are wrapped in `a` tags, and they should not be. We can easily fix this - I'll put one of our tech team on it.

walshbr:
Closes #1325


mdlincoln (approved):
Looks great 😄 🗡

PR #1380, opened 2019-07-15
Update creating-apis-with-python-and-flask.md

### LINKED ISSUE #1379: Typo in API's lesson
"line 416-417: function mentioned should be api_id, not api_root."

https://programminghistorian.org/en/lessons/creating-apis-with-python-and-flask

walshbr:
Closes #1379


walshbr:
Looks like this lesson hasn't been translated yet, so I don't think we need to tag any other teams. The error was in the text of the lesson and referred to a variable name from an earlier draft. But the code and everything still runs. Pretty minor but necessary fix.

acrymble (approved):
As long as you're confident in the change, this  is fine.

PR #1399, opened 2019-07-26
Update snippets.yml

alsalin:
adding myself back as EN managing editor


alsalin:
Trying to add myself back in as ME! Is this the edit to make for that now?

mdlincoln (approved):

PR #1410, opened 2019-08-08
Patching ME Blog Post

walshbr:
Trying to sort out gh-pages build error - it's hanging forever in the commit log even though the travis checks are passing, and we got a generic email about it.


acrymble (approved):

walshbr:
I'll merge this once the checks pass. @acrymble I'm not sure what the problem is since GithHub's build error emails are frustratingly vague, but if I had to guess it would be either the single quote in the title field (single quotes make it think it might be looking at a string that is never closed with another single quote) or the capital letter in the URL (it's probably going to lowercase that when building that site, which might make for weird file relationships).

PR #1413, opened 2019-08-13
Escape ampersand in gravity model lesson

### LINKED ISSUE #1412: RSS Error
"Hi, not sure if this is the best way for this kind of feedback, but I just recently found my way to the superb Programming Historian website, and wanted to subscribe to the RSS feed - but there seems to be some kind of error in the feed's syntax.

 

My RSS reader accepted the error as such, but the feed doesn't show any posts since January 2019 (and clearly you've published stuff since).
"

walshbr:
Closes #1412. The ampersand appears to be causing the RSS feed to not render in a valid manner.

### DIFF
--- a/_includes/feed-item.html
+++ b/_includes/feed-item.html
@@ -1,5 +1,5 @@
 <item>
-<title>{{ page.title }}</title>
+<title>{{ page.title | xml_escape }}</title>
 {% if page.abstract %}
   <description>{{ page.abstract | xml_escape }}</description>
 {% else %}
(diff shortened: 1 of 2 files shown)


walshbr:
I checked this, and we should be valid now. At least according to https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fdeploy-preview-1413--ph-dev.netlify.com%2Ffeed.xml

mdlincoln (changes requested):
how about we use `xml_escape` in the RSS feed template for all titles https://jekyllrb.com/docs/liquid/filters/

walshbr:
Ah gotcha - I didn't see that in the include. Thanks @mdlincoln

mdlincoln (approved):

mdlincoln:
🎊 📯 🦄 🐟

Issue #1347, opened 2019-06-21
Intro to Beautiful Soup lesson failed to work for me on Windows and Linux


weaverbel:
There are a few issues with this lesson.
1) Not always clear where one is meant to be - on the command line, in a Python IDE, in a Jupyter notebook, whatever. All the instructions seem to be written for a command line system, but that is not stated, and if you are not in that kind of system, then you need help on where you can run all of this. It is important to say what tools will be used and what environments *won't* work.
2) The first bit of code with `prettify` works fine but once you try the next bit, you get a warning about no HTML parser being specified and something about 'lxml' and so on and then nothing works after that.
3) There is also an error in the `print link` instruction. This only worked when I typed it as `print(link)` but I was not sure if that was what was actually meant. 
4) Building code iteratively is OK, but it is pretty confusingly laid out in this lesson and it isn't clear when to start again. Does `prettify` get chucked out altogether?
5) No advice on troubleshooting is provided, but given how flaky this code is, I think there needs to be *some* advice, even if only to circumvent the error messages I encountered.

I ran it on a Windows machine and then I ran it on a Linux machine running Ubuntu. Both times it all bombed out, and for the same reasons.

Never had this issue with other lessons so I am pretty disappointed today !

drjwbaker:
Thanks for your comments @weaverbel. Sorry you had some problems. On order:

1. The lesson does mention 'terminal' early on and this the implied environment, but I take the point about absolute clarity. It is medium difficulty lesson, so there are some assumptions baked in. Also, I'm not sure Jupyter was that much of thing when the lesson was written in 2012!
2. Okay. This seems like a problem we can fix. @jerielizabeth: are you able to take a look and suggest what is going on here?
3. As 2.
4. @weaverbel: are you able to specific which section of the tutorial you are referring to here?
5. Our tutorials don't tend to do this. They are walk-throughs of a process rather than if/but guides to an approach.

weaverbel:
Thanks @drjwbaker 
The tute starts with the full code so you could just go straight to that. But then it does a step by step thing so you can understand what you are doing as you go. You do the bit with `prettify` and then the next lot of code doesn't have `prettify` at all and it seems that bit is gone altogether. I wasn't sure if it needed to stay in or not as it wasn't in the full code at the start. I wanted to save the code for later so it would be good to clear that up.

[...]

PR #1415, opened 2019-08-16
updates to beautiful soup tutorial to address issues from #1347. 

jerielizabeth:
requires review. 

Attn @drjwbaker 

(and clearly I can no longer spell ... "embarrassing" typo.)

### DIFF
--- a/en/lessons/intro-to-beautiful-soup.md
+++ b/en/lessons/intro-to-beautiful-soup.md
@@ -81,6 +81,30 @@ sudo pip install beautifulsoup4
 
 {% include figure.html filename="https://imgs.xkcd.com/comics/sandwich.png" caption="The power of sudo: 'Sandwich' by XKCD" %}
 
+Additionally, you will need to install a "parser" for interpretting the HTML. To do so, run in the terminal:
+
+```
+pip install lxml
+```
+
+or 
+
+```
+sudo pip install lxml
+```
+
+Finally, so that this code works with either Python2 or Python3, you will need one helper library. Run in the terminal:
+
+```
+pip install future
+```
+
+or 
+
+```
+sudo pip install future
+```
+
 Application: Extracting names and URLs from an HTML page
 --------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


drjwbaker:
What type of machine do you recommend that I test this on?

jerielizabeth:
If you can confirm that it works on a Mac and with Linux that’d be great! The issue mentioned that it didn’t work well in windows but that may be a larger problem.

drjwbaker:
@jerielizabeth Tested on Mac. Two questions.

1. The final code block produces the attached error. Is this okay? 
Screenshot 2019-08-29 at 09 26 03

2. The final code block has lots of (what looks to me as a Python novice as) commented out code. Do we need all that?

drjwbaker:
I'll test on Linux when I'm on a Linux machine next week.

jerielizabeth:
1. Yeah ... I think they changed their website a bit since I first created the lesson and there are a few additional strings that should be deleted for it to work as it did originally. It is not a fatal error and I can fix it - it will just require adding a few more steps to the lesson. Hitting that balance between updating and rewriting again.

2. There is a lot of commented out code, to address the concern in the issue about code being removed. Not sure which is the less confusing way when writing an executable file ... happy to change back to removing, rather than commenting out.

drjwbaker (changes requested):
Confirm this works on Mac and Linux (I don't have a Windows machine for testing). On commented our text https://github.com/programminghistorian/jekyll/pull/1415#issuecomment-526931565, can I suggest a small revision that a) *briefly* explains what commented our code is (as this is an Medium difficult lesson and full explanation is not appropriate), and b) notes why keeping commented code in can be useful. Happy then to pull this to the main repo.

jerielizabeth:
@drjwbaker Added some brief language about the commenting.

drjwbaker (dismissed):

drjwbaker:
@programminghistorian/technical-team sorry to be a pain, but I've run out of talent trying to understand the Travis build fail here. I'm happy to merge this PR once the build fail is resolved. Ta!

walshbr:
```
- ./_site/en/lessons/transforming-xml-with-xsl.html
  *  External link http://www.tei-c.org/index.xml failed: 404 No error
```
Means that in the transforming-xml-with-xsl lesson there is a link to the TEI website that no longer resolves. Looking at the internet archive, it looks like that link used to go to the main index page of the TEI consortium. So I'd just update the link to direct to https://tei-c.org @drjwbaker.

drjwbaker:
@walshbr Huh. Didn't realise that a build fail on files not included in a PR would make Travis fail. Fixed now. Can someone please merge when the Travis checks are complete.

walshbr:
I think it was still failing because that lesson was translated, so the Spanish lesson also linked to the bad link. Let's let that run and then I'll merge if it comes through.

walshbr (approved):

drjwbaker:
Ta.

Issue #1089, opened 2018-11-17
"PH in English"?


mdlincoln:
Raised in #1088 

Should we rename _The Programming Historian_ to _The Programming Historian in English_? What are considerations for/against?

acrymble:
We'd have to look at the repercussions for the ISSN, the Wordcat entry, and generally for brand recognition. I'd be against it because I think the advantages are outweighed by possible headaches. I understand the merit of de-anglicizing the project.

drjwbaker:
Agree there are repercussions for listings. Though I think brand recognition would be unaffected: we'd still be PH, but with PH in E, en ES, and en FR flavours.

vgayolrs:
Next to the considerations on the ISSN, I think it should continue calling The Programming Historian. That was the name of the original project and it was the one that attracted us to try some first translations some time ago. Certainly the sites in Spanish and French will be able to create their own lessons - I hope more and more - but they are born as versions. Titling the page as The programming Historian in english seems like a version of some other project (El historiador programador, L'historien numérique...).

jenniferisasi:
When I mentioned that I liked the idea of adding "in English" to PH I was only thinking it in terms of the landing page, to make the 3 blocks more uniform, as it still has "The Programming Historian" unaltered in any of the versions on the top-left corner; does that make sense? 
I also think the repercussions for listings would be too big.

arojascastro:
Yes, I agree with @vgayolrs and @jenniferisasi

mdlincoln:
I do like the idea of making clear on the landing page that one is the PH in english, while still keeping the overall _Programming Historian_ title and not fundamentally introducing a new journal title. It sounds like we might all be in agreement on that.

This still raises a few questions:

1. Users will see a bit of a discrepancy between the apparent "title" of this journal _The Programming Historian in English_ that they click on the home page, and _The Programming Historian_ that shows up everywhere else in the English lessons.
2. When do we refer to _The Programming Historian_ in-line vs. language specific versions in places like the about pages, guideline pages, etc? Right now, for example, `/about` refers to "_The Programming Historian_" while `/apropos` refers to "_Programming Historian en français_"

One possible answer to 1) would be to have the title be something like "_The Programming Historian_ (original English version)". I'm not very sure about the term "original" as it may seem to assert primacy for the English version in a way we don't intend - on the other hand, though, it does correctly identify for users "where do I go to find the PH lessons I used three years ago?"

drjwbaker:
[...]

PR #1417, opened 2019-08-19
remove "initial" from "English version" on index

acrymble:
I've removed "The Initial" from "The Initial English Version" on the front page. That makes it sound like the English publication is more important.


mdlincoln:
There was not-insubstantial conversation about this when we made the choice to call it this: https://github.com/programminghistorian/jekyll/issues/1089#issuecomment-441440524

I don't think it's warranted to change it now unless a preponderance of the team has changed their views on it.

acrymble:
Ok, I'll withdraw it.

PR #1419, opened 2019-08-20
Adds note about escaping quotation marks in captions for figures

### LINKED ISSUE #1416: Escaping Quotation Marks in Figure Captions
Our author guidelines for figures - https://programminghistorian.org/en/author-guidelines#figures-and-images - do not mention escaping quotation marks. In other words, 

`{% include figure.html filename="special-paste-values.png" caption="Choose "Values" in the "Paste Special" menu." %}`

Will cause a jekyll build error, because it will assume the first quotation mark after Choose closes the caption. To have internal quotation marks in a parameter like this you have to escape them like so - 

`{% include figure.html filename="special-paste-values.png" caption="Choose \"Values\" in the \"Paste Special\" menu." %}`

Anyway, this is something we should address. Three times now we've had authors/editors confused because their lessons appeared to be breaking when the couldn't find any errors. In each case, the problem was with actually older lessons with issues like these that only subtly get flagged in the github web interface (they're a lot more obvious when working on the command line). We should also mention that punctuation marks _have_ to be inside the closing quotation mark.

I can take care of this - it would just entail adding two sentences or so to the author guidelines that will need to get translated.

walshbr:
Closes #1416

Needs to be translated into both Spanish and French. ~two sentences each.

### DIFF
--- a/en/author-guidelines.md
+++ b/en/author-guidelines.md
@@ -252,11 +252,11 @@ To insert an image in your text, use the following format:
 
 {% raw %}
 ``` markdown
-{% include figure.html filename="IMAGE-FILENAME" caption="YOUR CAPTION" %}
+{% include figure.html filename="IMAGE-FILENAME" caption="YOUR CAPTION USING \"ESCAPED\" QUOTES" %}
 ```
 {% endraw %}
 
-Images may not appear in previews of your lesson, but your editor will ensure they render properly when the lesson is published.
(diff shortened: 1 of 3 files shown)


fdlaramee:
Hi @walshbr -- French translation done. Awaiting Spanish translation to approve PR.

jenniferisasi (approved):
I'm approving the changes after translating the note about escaping quotation marks in captions.

PR #1445, opened 2019-09-10
remove Marie Puren from Team listing.

acrymble:
Marie Puren has informed us that she needs to step down from the project. We thank her for her hard work. This pull request removes her listing from the project team and puts an end date on her team membership.


amandavisconti (approved):

arojascastro:
Should not we remove also her bio? It is not necessary?

acrymble:
It may be for publications, if there are any.

PR #1470, opened 2019-09-28
Update lesson extracting-keywords from Python 2 to Python 3.

frederik-elwert:
This updates one lesson as discussed in #1460.


acrymble:
These changes are all very minor code or link changes so I'm happy for them to be merged. I'm not an expert on Python 3 though so I wouldn't be able to say with confidence that these changes were sufficient to make this fully Python 3 compliant. @mdlincoln do you  have a technical suggestion here? Should we (probably me) retry the code entirely before approving, as a workflow for these Python updates? If it becomes a substantial re-write we can trigger a re-review or consider retirement? I suspect it's mostly just minor syntax stuff.

mdlincoln:
@acrymble yes, it's a good idea for you to fully re-run all the code to confirm that it functions.

acrymble:
Thank you so much for this @frederik-elwert . I've re-tried the code and it works fine. I do have one query though. Do you have to explicitly run python3? Does this line need updating?

> You can now run the program you’ve written with the following command:
> 
> python extractKeywords.py

TO:

> You can now run the program you’ve written with the following command:
> 
> python3 extractKeywords.py

?

frederik-elwert:
That's a good point. I'm afraid this depends on the system and setup. On Ubuntu using the default Python, you're right, you'd have to run `python3`. With anaconda, it installs it's Python 3 as `python`. I'm currently on the road, but I'll try to think of something. Maybe a note in the general Python into lesson that we can point to.

acrymble:
I guess it doesn't matter in this case? I don't think you made any changes that wouldn't work for  Python 2, and int ime, this will resolve itself as fewer people have Python2 installed.

frederik-elwert:
You are right, doesn't matter in this case and does not need to block the merge. Good to keep in mind for further lessons, though.

acrymble (approved):

Issue #916, opened 2018-06-18
Patreon


drjwbaker:
Issue on [Patreon](https://www.patreon.com/) as a mechanism for accepting donations of #915.

- [x] Test [Patreon]
- [x] Add Patreon widget/button to site design. Info https://www.patreon.com/brand/guidelines. @walshbr working on this.
- [x] Agree donation tier(s)
- [x] Agree goal(s)
- [x] Agree text on page
- [x] Add to [Service Integrations](https://github.com/programminghistorian/jekyll/wiki/Service-Integrations)
- [x] Add bank/checkout details

drjwbaker:
Patreon's model is about monthly donations. Patreon runs a 'Rewards' system whereby patrons get things in exchange for the £$€ level at which they donate (see https://www.patreon.com/timsherratt for an eg).

I suggest we have :

- a minimum tier (£$€1 or more) where we thank people by name on the site
- a second tier (£$€5 or more) where we thank people by name on the site and send them some stickers
- a higher tier (?) where we thank people by name on the site and do something super nice for them (ideas?)

Given that Patreon does not support one-time donations (unless the patron donates then cancels subsequent monthly donations), I shall also explore PayPal as an option.

acrymble:
I reserved us the most logical page on this by the way. "Programming Historian".

drjwbaker:
Oops, I've just done that as well..

drjwbaker:
(at 'The Programming Historian')

drjwbaker:
Patreon has a goal based system:

> Earnings-based goals
> Your goals are based on how much you earn on Patreon.
> "When I reach $500 per month, I’ll start a special podcast series where I interview 1 patron every month."

I suggest we have one for copy-editing, based on [Society for Editors and Proofreaders](https://www.sfep.org.uk/resources/suggested-minimum-rates/) minimum costs, and a (conservative) estimate of 3 hours to copy edit each lesson, and ~2 new lessons each month (which I know is more than we get), one goal can be:

> $300 Alongside improving our community outreach and internationalisation and protecting the Programming Historian from changes to service agreements of those ‘free’ services upon which we rely, if we reach this goal we can ensure that all new lessons are copy-edited.

We may also choose to have a second goal for copy-editing all new translations, though given volume of translation this may need some thinking through.

Note: we can change these goals as we please (that is, they aren't binding)

drjwbaker:
Page mockup (preview before submitting for review with Patreon)

[...]

PR #1502, opened 2019-10-14
DO NOT MERGE patreon test

drjwbaker:
**DO NOT MERGE**

just checking what patreon looks like after the paypal test


drjwbaker:
Work coming out of https://github.com/programminghistorian/jekyll/issues/916 I think this isn't displaying properly on the preview because the Patreon account isn't live. Catch-22 et cetera :)

Issue #1503, opened 2019-10-14
Proposal for donation implementation


drjwbaker:
Following work on donation mechanisms https://github.com/programminghistorian/jekyll/issues/1457, membership https://github.com/programminghistorian/jekyll/issues/1405, and business models https://github.com/programminghistorian/jekyll/issues/1487 we propose to use this ticket to agree on implementing of donations to support our work. There are a number of strands to this work, so this is my attempt - on behalf of the Project Development Team - to pull together an actionable proposal.

1) We propose Patreon as the primary mechanism for ongoing donation, the benefit of which being that it provides a sustainable income. We propose to implement the Patreon tiers listed at https://github.com/programminghistorian/jekyll/issues/916#issuecomment-541396704. Once agreed **these are hard to change** (adjusting the tiers means cancelling ongoing donations), so we need to get this right. *This is all setup and ready to go.*
2) We propose Paypal as the primary mechanism for one off donations, the benefit of which is that Paypal is a widely used checkout mechanism. We will also accept bank transfer, cheque, and cash. *This is all setup and ready to go.*
3) We propose that parts of the site be updated to generate donations. These include:
	a) Update to https://programminghistorian.org/en/support-us listing all possible donation routes. *This is ready to go in English https://github.com/programminghistorian/jekyll/pull/1504 and will move to translation if/when agreed.*
	b) Paypal and Patreon buttons in the site footer. *This has been tested https://github.com/programminghistorian/jekyll/pull/1502*
	c) Some kind of banner, perhaps as a call out on all our lesson pages, where we can display donation info and our latest campaign. *This is currently the least specified part of the proposal.* We'd be grateful for input from @jenniferisasi on this, and how this might fit with comms (e.g. the twitter bot)

**Under our [Lazy Consensus model](https://github.com/programminghistorian/jekyll/wiki/Programming-Historian-Governance) I move that we discuss this at our next Editorial Board meeting https://github.com/programminghistorian/jekyll/issues/1483 with a view to receiving all feedback no later than 4 November 2019**.

Note, I think this affects everyone in some way, so please take a look. For example:

- @jenniferisasi: this feeds into the comms strategy
- @programminghistorian/technical-team: scoping design of *3c* above will involve technical time.
- @programminghistorian/french-team @programminghistorian/spanish-team @programminghistorian/english-team: editorial impacts here include workflows for copyediting https://github.com/programminghistorian/jekyll/issues/1381 using funds generated and translation labour.
[...]

PR #1520, opened 2019-10-25
Paypal in lesson header test

drjwbaker:
Testing paypal in lesson header per https://github.com/programminghistorian/jekyll/issues/1503


jenniferisasi:
How about a little coin in place of the signs? Like the little person and the little dock on the other two banners :) 

Looks like the other icons are from https://fontawesome.com/icons/lock (pulled it from the file code) so i tracked this icon: https://fontawesome.com/icons/coins?style=solid  
I also like the piggy-bank

walshbr:
I'll implement one of these in a second and ping you both when i have something on the branch.

walshbr:
Sorry this was taking forever to regenerate locally, which was making things slow. Check it now! The coin icon is not available in the version of font awesome we're using, and upgrading means we lose the old icons. So I went with the credit card. I also changed the text to "Support PH" because "Support the Programming Historian" is so many more characters than "Peer Reviewed" and "CC by 4.0" that there weren't going to be any good design solutions for it. Does that text work for you all? Could also go with with "Support us" or "Donate" @drjwbaker @jenniferisasi

jenniferisasi:
I like it! I would change Support PH to Support us, but it is not a big deal.

walshbr:
I'm fine with either one really, since I drew the line at character counts :) Do you have one you would prefer @drjwbaker? I note that in #1503 there was a question of ambiguity over who was being supported.

drjwbaker:
Yeah, 'Support Us' could mean the authors in this context.

jenniferisasi:
> Yeah, 'Support Us' could mean the authors in this context.

Ah, then Support PH it is.

walshbr:
In that case this should be good and just in need of translation, right @drjwbaker?

drjwbaker:
Thanks @walshbr & @jenniferisasi. Pinging @acrymble as I know you care about such things. Does this look good to you? If so, we'll push towards translation.

drjwbaker:
> In that case this should be good and just in need of translation, right @drjwbaker?

@acrymble appears to be happy, so yes just translation needed. 

@programminghistorian/spanish-team @rivaquiroga: see https://deploy-preview-1520--ph-dev.netlify.com/en/lessons/analyzing-documents-with-tfidf. Only translation needed is the phrase 'Support PH'. Is 'Apoyar PH' correct?

@programminghistorian/french-team @spapastamkou: see https://deploy-preview-1520--ph-dev.netlify.com/en/lessons/analyzing-documents-with-tfidf. Only translation needed is the phrase 'Support PH'. Is 'Soutenir PH' correct?

mdlincoln:
So this is what I'm seeing on the preview deploy:

Screen Shot 2019-11-04 at 8 57 26 AM

Are you sure the patreon script link is correctly added in?

_includes/footer.html
```diff
@@ -8,6 +8,16 @@
...
 </div>
+ 
+<div class="d-flex flex-wrap justify-content-center footer-head">
+ <form action="https://www.paypal.com/cgi-bin/webscr" method="post" target="_top">
+  <input type="hidden" name="cmd" value="_s-xclick" />
+  <input type="hidden" name="hosted_button_id" value="7BGHUZRVS4LYL" />
+  <input type="image" src="https://www.paypalobjects.com/en_GB/i/btn/btn_donate_LG.gif" border="0" name="submit" title="PayPal - The safer, easier way to pay online!" alt="Donate with PayPal button" />
+  <img alt="" border="0" src="https://www.paypal.com/en_GB/i/scr/pixel.gif" width="1" height="1" />
+</form>
+<a href="https://www.patreon.com/bePatron?u=11816614" data-patreon-widget-type="become-patron-button">Become a Patron!</a><script async src="https://c6.patreon.com/becomePatronButton.bundle.js"></script>
```
mdlincoln: confirm that this is all that's needed to display Patron button

mdlincoln (changes requested):

_includes/footer.html
drjwbaker: apologies. the footer buttons got left in after my initial playing. If we are going to keep them in (and no need not to) by saying "confirm that this is all that's needed to display Patron button" do you mean the changes you've made work?

mdlincoln: I have made no changes...?

I do not like the footer buttons - I think the lesson buttons are fine. If you intend to remove the footer buttons, then do so.

fdlaramee:
@drjwbaker Yes, "Soutenir PH" is appropriate. Thanks!

drjwbaker:
@walshbr I think I'll added the snippet thingy correctly..

_includes/footer.html
drjwbaker: Sorry. I didn't understand your first message. I'll remove the footers. Aren't my favourite either.

walshbr:
Yep that's correct use of the snippets @drjwbaker 🎉  That portion looks good to me in the preview.

mdlincoln (changes requested):
Right now, the links only point to the English "support us" page, no matter which language page you're coming from. We should have the links point dynamically to the correct language page by relying on the existing snippet here: https://github.com/programminghistorian/jekyll/blob/gh-pages/_data/snippets.yml#L143

the `href` part of the link would need to be updated to read like `https://programminghistorian.org/{{ page.lang }}/{{ site.data.snippets.menu-contribute-support[page.lang].link}}`

drjwbaker:
Good catch @mdlincoln. I think I've done it right. Though I'm not sure if (if at all?) spaces between curly brackets do anything.

drjwbaker:
> Yep that's correct use of the snippets @drjwbaker 🎉 That portion looks good to me in the preview.

Shock horror.

drjwbaker:
Did it wrong. Let me try again..

mdlincoln (approved):

drjwbaker:
Sorry. Me again. How do we use the snippets to make the link go to the correct language sub-section of a page, rather than just the correct language page? So https://programminghistorian.org/es/apoyanos#donaciones rather than https://programminghistorian.org/es/apoyanos for https://deploy-preview-1520--ph-dev.netlify.com/es/lecciones/analisis-temporal-red.

drjwbaker:
pinging @programminghistorian/technical-team

walshbr:
Hmm…I think you would need to add a new snippet for the subsection you're interested in so the links would become something like this -

https://programminghistorian.org/{{ page.lang }}/{{ site.data.snippets.menu-contribute-support[page.lang].link}}#{{ site.data.snippets.menu-contribute-support[page.lang].subsection }}

Though note that the trailing piece after the slash would be - the actual structure following the slash would depend on what you name the snippet / where it is / etc.. You'd need to have some way to store the translated name of the subheader in a snippet so that you can call it in the link by referencing the page language.

PR #1522, opened 2019-10-31
Update Intro to BeautifulSoup for Python3

frederik-elwert:
Updated this lesson to be compatible with Python3. The code was already Py3 ready, so I basically removed the future imports.

Contributes to #1460.


mdlincoln:
@acrymble once you review this, I will force-merge, since the only failure is a 404 link that's been fixed upstream already.

acrymble:
I am not the author of this lesson, so I don't think I should be reviewing it. I think this falls under @svmelton as Managing Editor.

svmelton (approved):

PR #1536, opened 2019-11-04
Create beginners guide to twitter data

spapastamkou:
I am raising this pull request for Beginner's Guide to Twitter Data lesson. All files are in this branch so if there are any problems we can work them out from here.


walshbr:
@spapastamkou and @svmelton - the spacing wasn't quite uniform for the last few authors - that was another issue keeping the error message from being very helpful. Now these are the issues (once the checks finally stop running) - 

  - [x] 'date' is missing.
  - [x] '["data-manipulation"]' is not a valid lesson topic.
  - [x] '["api"]' is not a valid lesson topic.

walshbr:
Besides just adding a date, I think the topics don't need brackets around them if they're on their own hyphenated lines. If that doesn't work lmk.

spapastamkou:
thnx @walshbr for the help; tried to fix the spacing issue in ```ph-authors```` but you had done it already:-) 
The brackets are ok for the topics, but the hyphens are not, so I just fixed this. Let's see now.

svmelton (approved):

spapastamkou:
thank you @svmelton, seems we have a published lesson now!

PR #1553, opened 2019-11-16
direct link to patreon

acrymble:
I've changed the link in the 'support PH' at the top of each lesson to point directly to Patreon. From a UX standpoint we need to get people to actionable pages asap.

### DIFF
--- a/_layouts/lesson.html
+++ b/_layouts/lesson.html
@@ -56,7 +56,7 @@ All lesson metadata and alerts should follow the convention of pulling appropria
                     4.0</a></p>
               </div>
               <div class="donate mr-5">
-                <p><a href="https://programminghistorian.org{{ site.data.snippets.menu-contribute-support-donate[page.lang].link}}"><i class="fas fa-credit-card"></i> {{ site.data.snippets.donate[page.lang] }}</a></p>
+                <p><a href="https://www.patreon.com/theprogramminghistorian"><i class="fas fa-credit-card"></i> {{ site.data.snippets.donate[page.lang] }}</a></p>
               </div>
             </div>
           </div>


drjwbaker:
But we have other ways of accepting payment.

acrymble:
Yes but no one has used them to my knowledge. We need to focus on building our subscribers by making it as visible and easy as possible

mdlincoln:
@acrymble Where is this unilateral change coming from? @drjwbaker and the rest of the team have spent quite a while discussing and designing this flow, including creating language-specific donation pages that you're now un-linking. Why did you not bring this up during the long review process? Team-based decision making can't work if you're immediately going to propose changes _after_ something has _just_ been implemented.

drjwbaker:
Also @acrymble, it has only been a few weeks with minimal comms re donating.

acrymble:
@mdlincoln there are several separate issues here.

1) I'm not proposing we delete the support us page or its new translations. I'm suggesting we cut out a step for people who click on the 'support us' link on a lesson page. This gets them to an actionable page (and a much more visually stimulating one) much quicker. I believe this will increase our chances for attracting donors. We are 10% to our target of $500 per month. According to Google Analytics, only about 4 people per day visited the 'Support us' page in the week since it's been updated. That might include us taking a look at it. We're going to have to keep tinkering this until we find out what works best.

2) This is not a unilateral decision but a suggestion made to the appropriate team member via a  pull request, and based upon our agreed workflows of team based (not board-based) decision making as decided at our last meeting and circulated via the meeting minutes. (10. Internal Communications #1491  * Agreed: Teams are empowered to get on with decisions within their remit, and raise issues for the group when they feel issues are substantive enough to need team input.) This is part of a conversation that took place via email between @drjwbaker and I as members of the Development Team, under which Patreon falls.

3) We are going to lose good people (and recently did) if we can't afford to offer fair compensation for people on precarious contracts or without jobs.

If you're all more comfortable, I have set up  Google Analytics for Patreon. Let's A/B test these two pages for a few weeks each, and see if it makes a difference one way or another.

acrymble:
Closing this to move back to conversation on tickets before additional changes.

PR #1584, opened 2019-12-08
Moving files for Jupyter Notebooks lesson

svmelton:
Moving files to prepare for Jupyter Notebooks lesson publication


en/lessons/jupyter-notebooks.md
```diff
@@ -0,0 +1,391 @@
...
+
+Next, you need to figure out how to do the conversion. Searching for relevant terms may lead you to [this StackOverflow discussion](https://stackoverflow.com/questions/2265357/parse-date-string-and-change-format), and the first answer involves using the *datetime* Python module. As a first step, you need to import *datetime*, using a code cell. You also know that your input file is a CSV, so you should import the *csv* module as well.
+
+To add a new cell, click the plus button <i class="fa fa-plus"></i> in the toolbar (or use the *esc + b* keyboard shortcut). This will create a new code cell below the cell that's currently selected. Create a new code cell, and paste in the following code to import a Python module:
+
+```
```
mdlincoln: In all the code blocks that contain python, we can enable code highlighting by appending `py` after the three backticks

```
``` py
```

(make sure to include the space between the 3rd backtick and the `py`)

This will help make the code chunks, especially the very long ones, much more readable - and it will better match the formatting that students will see in the Jupyter interface, too!

mdlincoln (changes requested):

en/lessons/jupyter-notebooks.md
svmelton: Oh, very cool! I just added it to the python codeblocks.

walshbr:
@mdlincoln is your sense that the editor of the individual lesson should be the one approving the PR by the ME for the actual publication? Might be good to have someone else do it as a general rule, as was the case with your comments here. Seems useful to get more eyes on it, and I hadn't thought about the Python code highlighting.

mdlincoln:
@walshbr hm, indeed, our editor guidelines don't actually specify.

I like the idea of tagging the language team rather than specifying one individual person to finally review and approve. I don't think it necessarily makes sense to prohibit the lesson editor from doing the approval review, since the ME is already supposed to be that pair of fresh eyes. But 🤷‍♂

mdlincoln (approved):

Issue #1587, opened 2019-12-09
Error in Exploring and Analyzing Network Data with Python 


svmelton:
A reader reported the following error via email:
`for n in G.nodes(): # Loop through every node, in our data "n" will be the name of the person
    print(n, G.node[n]['birth_year']) # Access every node by its name, and then by the attribute "birth_year"
`

Should be:
`for n in G.nodes(): # Loop through every node, in our data "n" will be the name of the person
    print(n, G.nodes[n]['birth_year']) # Access every node by its name, and then by the attribute "birth_year"
`

I'll submit a PR.

walshbr:
Hmm…wonder if the API for networkx might have changed, as that syntax shows up a few times in the lesson - "G.node[" shows up several times times. And all the code blocks definitely ran when I was editing the thing back when it live. Let me know if you'd rather me do a PR, @svmelton. Or happy to just review yours - all the same to me. Thanks! Good catch.

svmelton:
If you've got the time to do a PR, that would be great!

walshbr:
Actually @svmelton - it looks like updating that syntax exposed another issue that is indeed from ways that API changed in a newer version in the last couple years. So I think it won't be as easy as just updating that one piece of syntax. 

@jrladd - can you take a look at this and let me know what you'd like to do? One option might be to freeze the version number on the lesson, or we could update it for the more recent networkx version. Besides the piece flagged here, the line here - 

"communities = community.best_partition(G)" 

is throwing an error because "edges_iter" was removed in the more recent version of networkx. Let me know what you think!

svmelton:
@walshbr @jrladd I had to run the `pip3 install python-louvain --upgrade` command and install numpy to get that section to work.

I am running into issues on the eigenvector section, however—I'm getting a keyerror on eigenvector.

jrladd:
Thanks @svmelton @walshbr for looking into this. These difficulties are definitely because NetworkX is changing all the time. I'd prefer to update rather than leave it with an old version, but we should probably note the version this time in case this happens again. It's unlikely we'll be able to keep up with NetworkX forever!

For the nodes issue, updating to `G.nodes[n]` in all the places it occurs makes perfect sense. @svmelton where exactly is the KeyError coming up in the eigenvector section? Is it not finding what it expects in `eigenvector_dict`? I suspect this is also due to a change in the NetworkX API, but I'm wondering why it isn't affecting betweenness. It may be that I forgot something the last time I updated the code.

[...]

PR #1626, opened 2020-01-17
Updates to Exploring and Analyzing Network Data with Python

jrladd:
In response to #1587 I'm sharing a set of updates that should resolve those errors. These updates include:

- Setting a stable install version in the tutorial, so that all future users will install NetworkX v2.4 (the current latest stable release)
- Changing all instances of `G.node[]` to `G.nodes[]` in accordance with the latest NetworkX syntax
- Eliminating the python-louvain module in favor of the new built-in NetworkX modularity method (this one required the most changes to the lesson, but I think the overall effect is simpler and more straightforward)

I wanted to give you all (@svmelton @walshbr @mdlincoln) a chance to look this over. Thanks for pointing out these issues! Let me know if these solutions will work and if there's anything else I can do.


walshbr:
@svmelton - I've run everything and confirmed it all works. Since you were also taking a look at this, did you want to take a run at things? Otherwise this can be approved and merged.

jrladd:
Thanks @walshbr! Also, I just realized that I never removed the Alert text, so you may want to do that as part of this update.

svmelton (approved):

PR #1696, opened 2020-03-02
add new ES ombudsperson

rivaquiroga:
Changes involve not only adding Silvia's avatar + bio in `ph_authors.yml`, but also editing `guia_editor.md` and her name in the lesson she wrote, as she prefer using both of her last names.

Closes ##1443


rivaquiroga:
@acrymble, it seems that the new reordering of the project team page has as a consequence that only editors can appear under each publication title. If I don't add the `spanish` tag to Silvia in`ph_authors.yml`, she is not added to the page. The problem is that it adds the label "editor", and she doesn't has that role. 
I also just realized that Amanda Visconti doesn't appear as Ombudsperson in the Project Team page.

jenniferisasi:
good catch, Riva! I will wait for @acrymble to check on that if he can before proceeding to approve the changes.

acrymble:
Can you open a ticket for us to add an "Ombudsperson" section? Then we don't have to hold up the process.

acrymble (approved):

Issue #1698, opened 2020-03-04
Split translation concordance into lessons vs. other pages sections


jenniferisasi:
As per conversation in [238 in the ph-submissions](https://github.com/programminghistorian/ph-submissions/issues/238) repo I will attempt to update the [concordance list](https://programminghistorian.org/translation-concordance) we have to show the state of translations of lessons to make it easier for potential translators to decide on tasks. 

Matthew Lincoln recommended: 
If you use the where liquid filter you can isolate the pages that are lessons, and use those with the existing code.

Since I am on transition into a new job, I will need some time to work on this, but it should be done **by the end of the month** (March).

PR #1705, opened 2020-03-08
Split translation concordance into lessons vs. other pages sections 

jenniferisasi:
My first attempt at organizing this page to reflect the progress of translated content per #1698. 

Will probably take a few attempts in my learning process.

### Checklist

- [ ] Assign at least one individual or team to "Reviewers"
  - [ ] if the text needs to be translated, assign the relevant language team(s) as "Reviewers" and tag both the team as well as the managing edtor in your PR. Please follow the [translation request guidelines](https://github.com/programminghistorian/jekyll/wiki/Requesting-Translation-Guidelines) when writing your PR description
- [ ] [Ensure the Travis CI checks pass](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#4-check-your-pr-status)
- [ ] [Check the live preview of your PR on Netlify](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#5-preview-how-your-pr-looks-when-built-into-html)
- [ ] If this PR closes an open issue, add the phrase `Closes #ISSUENUMBER` to the description above
- [ ] Duplicate this page and add it to the other languages' folder. I will translate the only sentence myself. 

*If you are having difficulty fixing Travis errors, first consult <https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions> carefully, especially ["Common Travis Errors"](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#common-travis-errors). Then contact the technical team if you need further help.*

### DIFF
--- a/translation-concordance.md
+++ b/translation-concordance.md
@@ -3,9 +3,9 @@ layout: blank
 title: Translation Concordance
 ---
 
-An automatically-generated list of lessons translation relationships across our publications.
+An automatically-generated list of page translation relationships across our publications.
 
-{% assign original_pages = site.pages | where: "layout", "lesson" | where_exp: "item", "item.retired != 'false'" %}
+{% assign original_pages = site.pages | where_exp: "item", "item.name != 'redirect.html'" | where_exp: "item", "item.name != 'redirects.json'" | where_exp: "item", "item.name != 'index.md'" | where_exp: "item", "item.original == nil" %}
 
 <table>
   <tr>{% for l in site.data.snippets.language-list %}
@@ -18,6 +18,27 @@ An automatically-generated list of lessons translation relationships across our 
   <tr>
     {% for l in site.data.snippets.language-list %}
     {% assign sp = page_versions | where: "lang", l | first %}
+    {% assign where_exp: "item", "item.layout == 'lesson' %}
+    <td><a href="{{sp.url}}">{{ sp.title }}</a></td>
+    {% endfor %}
+  </tr>
+{% endfor %}
+</table>
+
+{% assign original_pages = site.pages | where_exp: "item", "item.name != 'redirect.html'" | where_exp: "item", "item.name != 'redirects.json'" | where_exp: "item", "item.name != 'index.md'" | where_exp: "item", "item.original == nil" %}
+
+<table>
+  <tr>{% for l in site.data.snippets.language-list %}
+    <th>{{ l }}</th>{% endfor %}
+  </tr>
+  {% for p in original_pages %}
+  {% assign reverse_p = p.name | split: "." | first %}
+  {% assign translated_pages = site.pages | where: "original", reverse_p %}
+  {% assign page_versions =  p | concat: translated_pages %}
+  <tr>
+    {% for l in site.data.snippets.language-list %}
+    {% assign sp = page_versions | where: "lang", l | first %}
+    {% assign where_exp: "item", "item.layout != 'lesson' %}
     <td><a href="{{sp.url}}">{{ sp.title }}</a></td>
     {% endfor %}
   </tr>
--- a/translation-concordance.md
+++ b/translation-concordance.md
@@ -24,7 +24,7 @@ An automatically-generated list of page translation relationships across our pub
 {% endfor %}
 </table>
 
-{% assign original_pages = site.pages | where_exp: "item", "item.name != 'redirect.html'" | where_exp: "item", "item.name != 'redirects.json'" | where_exp: "item", "item.name != 'index.md'" | where_exp: "item", "item.original == nil" | where_exp: "item", "item.layout != 'lesson'" %}
+{% assign original_pages = site.pages | where_exp: "item", "item.name != 'redirect.html'" | where_exp: "item", "item.name != 'redirects.json'" | where_exp: "item", "item.name != 'index.md'" | where_exp: "item", "item.original == nil" | where_exp: "item", "item.layout != 'lesson'" | where_exp: "item", "item.layout != 'post'" %}
 
 <table>
   <tr>{% for l in site.data.snippets.language-list %}
--- a/translation-concordance.md
+++ b/translation-concordance.md
@@ -5,7 +5,7 @@ title: Translation Concordance
 
 An automatically-generated list of lessons translation relationships across our publications.
 
-{% assign original_pages = site.pages | where_exp: "item", "item.name != 'redirect.html'" | where_exp: "item", "item.name != 'redirects.json'" | where_exp: "item", "item.name != 'index.md'" | where_exp: "item", "item.original == nil" %}
+{% assign original_pages = site.pages | where: "layout", "lesson" %}
 
 <table>
   <tr>{% for l in site.data.snippets.language-list %}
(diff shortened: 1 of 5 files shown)


jenniferisasi:
My understanding is that the original code 
`{% assign original_pages = site.pages | where_exp: "item", "item.name != 'redirect.html'" | where_exp: "item", "item.name != 'redirects.json'" | where_exp: "item", "item.name != 'index.md'" | where_exp: "item", "item.original == nil" %}`
was calling to all the pages that where not (!=) a redirect, an index or an empty page. 

So I changed it to a simple `where`:
`{{ site.pages | where: "layout", "lesson" }}`
to pick only pages that are lessons by checking their layout. 

That seemed to work as now only lessons are showing BUT I need to change something in the next block of code because some are duplicate, out of place, by themselves...

Retired lessons are also showing.

translation-concordance.md
```diff
@@ -3,9 +3,9 @@ layout: blank
 title: Translation Concordance
 ---
 
-An automatically-generated list of page translation relationships across this site.
+An automatically-generated list of lessons translation relationships across our publications.
 
-{% assign original_pages = site.pages | where_exp: "item", "item.name != 'redirect.html'" | where_exp: "item", "item.name != 'redirects.json'" | where_exp: "item", "item.name != 'index.md'" | where_exp: "item", "item.original == nil" %}
```
mdlincoln, line 8: Thanks for starting to take this on!

First off, we have to keep all the complex existing filters because that's what prevents the repeat / disjointed table you're currently seeing. In particular the `where_exp: "item", "item.original == nil"` makes sure that each row in the table is fixed to one "original" page (no matter what language the original is).

It would be useful for the concordance to show TWO tables - one for lessons, and one for all other pages. So I'd start by duplicating all the **original** code (before you made these modifications) from `{% assign original_pages...` all the way through the end `</table>` tag.

Then, we'll need to modify both of the `assign` liquid calls. For the one that should only include lessons, I'd add a filter at the end that does `where_exp: "item", "item.layout == 'lesson'"`. For the one that should show all non-lesson pages, `where_exp: "item", "item.layout != 'lesson'"`

Try to get that working first, then we'll double-check how to filter out retired lessons - although I think we might actually want to keep retired lessons in this list since they're still valid pages on our site, just not ones that we list up front in the lessons directory.

Sorry, that's a lot for just one comment - let me know if it makes sense or if I can clarify anything in order for you to try and make the changes I suggested.

mdlincoln (changes requested):

translation-concordance.md
jenniferisasi: Thank you, @mdlincoln! I didn't know (as you can guess) that I needed to keep the where_exp code. If we think retired lessons should start, we keep them. 

I am short on Internet connection so I'll probably take a few more days to try to make these changes.

mdlincoln:
@jenniferisasi I I do not think we will have time to discuss this tomorrow during #1722. When you can, please provide a status update though. Do you need any help to finish this PR?

jenniferisasi:
> @jenniferisasi I I do not think we will have time to discuss this tomorrow during #1722. When you can, please provide a status update though. Do you need any help to finish this PR?

Sorry, I haven't had time to look at this again [facepalm emoji]. I will try during during the week though I might need some more time.

jenniferisasi:
I understand the error on liquid syntax being wrong. As history shows, I thought I could do a pipe on the last assign to add the new filter, but it took everything out... so in order not to rever to it, I leave it like this for now.

mdlincoln:
Thanks for putting in more updates @jenniferisasi ! If it's any consolation, this is advanced jekyll/liquid code that you're editing, so that's why it's difficult.

I'll add some inline comments to your latest commit - don't give up yet 😸

translation-concordance.md
```diff
@@ -18,6 +18,27 @@ An automatically-generated list of page translation relationships across this si
...
+    <th>{{ l }}</th>{% endfor %}
+  </tr>
+  {% for p in original_pages %}
+  {% assign reverse_p = p.name | split: "." | first %}
+  {% assign translated_pages = site.pages | where: "original", reverse_p %}
+  {% assign page_versions =  p | concat: translated_pages %}
+  <tr>
+    {% for l in site.data.snippets.language-list %}
+    {% assign sp = page_versions | where: "lang", l | first %} 
+    {% assign where_exp: "item", "item.layout != 'lesson'" %}
```
mdlincoln: This is where things go off the rails - it's not where the `where_exp: "item", "item.layout != 'lesson'"` belongs. You want to put it instead up in the `assign` statement on line 28, after `where_exp: "item", "item.original == nil"`.

Everything else inside the `<table>...</table>` tags ought to stay the same. What we're trying to do is create a new version of the `assign` statement in the original code except that it also filters to just lessons (or non-lessons, depending on which table we're making), and then using the exact same table code to loop through.

mdlincoln (changes requested):

jenniferisasi:
Hmm filtering out by `post` didn't work... and I guess it is because it is not actually showing the blog `post` but the entire blog... have to think about it.

mdlincoln:
Oh, yes, interesting. Those pages are automatically created by Jekyll pagination, which we added on in the past month or so. Off the top of my head, I don't know how to exclude them from the concordance but the answer will probably be somewhere in adding a yaml attribute to the blog list page (it could be an arbitrary key-value combo), and then trying to filter out that combo.

That said, congrats on getting the core part of the job finished! We have two tables, one with lessons, the other wihout lessons.

jenniferisasi:
Another thing I realized exploring the Netlify version, is that the page doesn't transfer in ES | FR (language menu); however the page is on the About menu in the 3 languages, if you click, it takes back to the EN version page (which is not terrible). My best bet is that this happens because the file lives outside the es, fr, en, pt folders, on the main repo. 

In exploring our file organization again...

@mdlincoln will it work if I duplicate & translate the file I just edited, and put a copy in each of the languages folder, without needing any other code change somewhere else?

mdlincoln:
You are correct that it is only on the en page now. It should work if you copy it, but let's leave that as a follow up task once the code for this page is settled

mdlincoln:
@jenniferisasi take a look at the couple of commits I just pushed to see if they make sense to you.

mdlincoln (approved):

jenniferisasi:
@mdlincoln just checking the additional changes you made to the code. I believe I understand the skip_concordance add on. Thanks!

There are a couple of things that show up so I will leave this page open on my browser to look at them when I get a couple of minutes.

jenniferisasi:
@mdlincoln I had the time to looks at all the changes and I think this "issue" is the perfect example for me to confirm that I understand the fix you did: 
These two pages are showing; whereas one is useful in the sense of a guide, the other one (in Spanish) is the readme.md of a set of assets/files for a lesson (doesn't belong here). Both files are missing YAML information, so the filters cannot decide to not show them (like the blog before). 

In these cases (thinking of .md files we might add in the future for some reason) would a yaml with `skip_concordance: true` just suffice? 

Screen Shot 2020-05-12 at 7 26 09 PM

mdlincoln:
>In these cases (thinking of .md files we might add in the future for some reason) would a yaml with skip_concordance: true just suffice?

yes

PR #1974, opened 2021-01-11
Tweak translation concordance to avoid files with no titles

### LINKED ISSUE #1972: Translation Concordance Page Bugs
While working on my MLA presentation for ProgHist I noticed that there are some issues with the translation concordance page - https://programminghistorian.org/translation-concordance. There are a bunch of blank rows, some pages are showing up in Spanish under the English column, and we'll want to make sure PT shows up there after that site launches. I've assigned myself for now, but if any other @programminghistorian/technical-team person has time and wants to take a look go for it. This is a screenshot of the bottom of the page - 
Screen Shot 2021-01-08 at 2 38 01 PM

walshbr:
Closes #1972 

The translation concordance page had a bunch of blank rows in its "other pages" table. These were being caused by a series of assets and image files that were making it into the logic of the loop. I added a little logic to the loop for it to avoid pages that have no titles. @jenniferisasi, can you take a look and make sure that this didn't break what you had there before? It's a little brute force, but I think it does what we need.

### DIFF
--- a/translation-concordance.md
+++ b/translation-concordance.md
@@ -26,7 +26,7 @@ An automatically-generated list of page translation relationships across our pub
 {% endfor %}
 </table>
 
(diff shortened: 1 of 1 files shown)


jenniferisasi:
Look Travis is showing the twitter broken links again!

jenniferisasi (approved):
It works perfectly!! And actually makes a lot of sense to exclude pages without titles.

PR #2208, opened 2021-07-22
Publicar introducao omeka net

DanielAlvesLABDH:
*Publicar tradução da lição "Up and Running with Omeka.net"*

### Checklist

- [x] Assign at least one individual or team to "Reviewers"
- [x] [Ensure the status checks pass](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#4-check-your-pr-status)
- [x] [Check the live preview of your PR on Netlify](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#5-preview-how-your-pr-looks-when-built-into-html)
- [ ] If this PR closes an open issue, add the phrase `Closes #ISSUENUMBER` to the description above

*If you are having difficulty fixing build errors, first consult <https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions> carefully, especially ["Common Build Errors"](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#common-build-errors). Then contact the technical team if you need further help.*


DanielAlvesLABDH:
Could someone from the @programminghistorian/technical-team help me on this? It's probably a silly question but when we use images in the lesson that are different from the original ones, we should put those new images in the original (in this case EN) folder or should we create a new folder, with the name of the translated lesson? I took the second option and the build failed. So, should I move the images for the original (EN) folder? Thanks in advance

spapastamkou:
We use the folder of the original (you can see [the example](https://github.com/programminghistorian/jekyll/tree/gh-pages/images/cleaning-data-with-openrefine) of the lesson for cleaning data with Openrefine where the ES and FR images are in the folder of the original)

DanielAlvesLABDH:
Thank you @spapastamkou !

acrymble:
> We use the folder of the original (you can see [the example](https://github.com/programminghistorian/jekyll/tree/gh-pages/images/cleaning-data-with-openrefine) of the lesson for cleaning data with Openrefine where the ES and FR images are in the folder of the original)

This has me worried about problems down the line. If a lesson is retired in one language that could break image paths. Each should be treated as a distinct lesson, shouldn't it?

DanielAlvesLABDH:
That was my thought @acrymble. For me it was logical that the images should be also in the translated folder. But again that could result in duplicating files, if we use the same images...

acrymble:
@DanielAlvesLABDH I think duplication is safer. You don't want someone updating a file for the English lesson because the interface changes, and in the process make a change that doesn't make any sense for the PT lesson.

There's a reason why we insist that datasets be kept on our servers, and it's to prevent problems like this. Can we get a sense of how big this problem might be across the various teams? We might have to un-do it all to ensure sustainability.

DanielAlvesLABDH:
If the duplication is not a problem, than I'm in favor of the change. But it's going to be some good amount of work for this, I guess.

walshbr:
Doing that would increase build times and take up more space in the repository (that's probably not that big a deal, though, unless you start adding lots more languages. But note that @ZoeLeBlanc is actively looking at how to reduce build times. It might be necessary but just wanted to note the effects it would have. I suppose the slightly more burdensome thing would be to make it a policy to give more distinctive filenames that tie into a lesson version number or something like that and be careful not to overwrite old images with new ones. But @ZoeLeBlanc would have a better sense how big of a deal the increase in load time would be if all the translated lessons got duplicate image folders.

drjwbaker:
On builds, this month we've hit 50% of netlify in the first 11 days of our plan. I think we're looking at busting out of free plan almost every month ($7 per month per month after that isn't a lot, and we can afford it, but just noting).

walshbr:
Just for context - I don't think @ZoeLeBlanc and I were primarily concerned about the finances for Netlify with build time (though that's of course an issue - thanks for sharing the stats!). The Jekyll site (much like the Scholars' Lab's Jekyll site) is just getting way more complex than most Jekyll sites, and it makes it difficult to actually do any kind of technical work on it. Rather than waiting a couple seconds for the site to regenerate like most Jekyll projects, ours takes a couple minutes (you've all experienced that with how long it takes for the pull requests to build, actually). That meant that the accessibility work I was doing took days rather than hours, just because I had to go do other stuff while I waited for it to spin up a change to a single HTML tag over the course of several minutes. But you might already be aware of all this - I haven't been at some of the recent conversations about rebuilding the thing! Just offering my thoughts.

drjwbaker:
Thanks for the context @walshbr. @ZoeLeBlanc: if there is a way to spend £$€ to bring down the build time to save us time, then please don't rule that out (that is, the spend we have isn't a problem, more than I'm noticing our netlify build use going up)

ZoeLeBlanc:
Totally 💯 agree with everything @walshbr wrote above about build times, but just wanted to add that from a minimal computing perspective, having the same images twice is also adding a bit of unnecessary overhead for loading the website. I think for now we could test making the change and see how that impacts build times, but then longterm we might want to reconsider how we organize assets for lessons when we start looking at other frameworks for our tech infrastructure (probably won't happen until Winter earliest though).

joanacvp (approved):
@DanielAlvesLABDH Fiz pequenas mudanças. Está pronto a ser publicado.

PR #2731, opened 2022-11-03
Issue 2707 edits

### LINKED ISSUE #2707: Issues with "Scalable Reading of Structured Data"
While translating "Scalable Reading of Structured Data" we found some issues related to the plots shown in the lesson. 

## Figure 1
- The number of tweets is hardcoded in the plot with `caption = "Total number of tweets: 2.413"`. The problem is that the plot included in the lesson says the number of tweets is 2435. This mismatch could be potentially confusing for readers, especially because if we add the actual number of tweets from column "n" of the dataframe shown in the lesson, we get 2432 tweets.

- The code suggests changing the axis breaks with `scale_y_continuous(breaks = seq(0, 400, by = 50))`, but the plot included was not created using the code provided in the lesson (there are no ticks or values on axis "y"):

figure1-o

 
This could be potentially confusing for readers. 
(Actually, we decided to remove that line of code in the Spanish translation because if the number of tweets that the reader gets is different, you can end with a plot with a "y" axis like this one:

figure1

 
## Figure 2

The plot says the number of tweets is 2435, but in the output shown in previous lines, the total number of tweets is 2432.

## Figure 3

The description above the plot says: _The visualisation looks a lot like the previous bar charts, but the difference here is `facet_wrap`, which creates three bar charts for each type of interaction._ The problem is that the image only shows one bar plot for one type of interaction. In this case, it is not a mismatch between code and plot. The reason is that in the dataframe `interaction` (the one used for creating the plot), the column “interaction” only has one value (the string "favorite_count"). It was intentionally created that way. That is why `facet_wrap` only creates one plot. 
 
This is the code used in the lesson for creating `interactions`

``` 
interactions <- sesamestreet_data %>%
  group_by(verified) %>%
  summarise(mean = mean(favorite_count)) %>%
  mutate(interaction = "favorite_count")
```

And this is the content of that object:

``` 
  # A tibble: 2 × 3
  verified    mean interaction  
  <lgl>      <dbl> <chr>        
1 FALSE      0.892 favorite_count
2 TRUE     114     favorite_count
```
 
It is not clear to us what the three bar charts were supposed to be depicting. To have three different plots using `facet_wrap`, the variable called within that function should have three different values. But the column “interaction” was intentionally created containing the same value in all rows: `mutate(interaction = "favorite_count")`.
That section ([Interaction with verified versus non-verified account](https://programminghistorian.org/en/lessons/scalable-reading-of-structured-data#interaction-with-verified-versus-non-verified-accounts)) starts getting a little confusing from this line: _In this next step, you add the result of the calculation above to a dataframe. Use a new column headed “interaction” to specify that it is “favorite_count”_. The result of the calculation is already a dataframe. The next step (`mutate(interaction = "favorite_count")`) just adds a column to that same dataframe. Is there a step missing?

anisa-hawes:
This PR resolves Issues identified in Lesson Maintenance report #2707.

Closes #2707 

### Checklist

- [x] Assign at least one individual or team to "Reviewers"
  - [ ] ~~if the text needs to be translated, assign the relevant language team(s) as "Reviewers" and tag both the team as well as the managing edtor in your PR. Please follow the [translation request guidelines](https://github.com/programminghistorian/jekyll/wiki/Requesting-Translation-Guidelines) when writing your PR description~~
- [x] [Ensure the status checks pass](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#4-check-your-pr-status)
- [x] [Check the live preview of your PR on Netlify](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#5-preview-how-your-pr-looks-when-built-into-html)
- [x] If this PR closes an open issue, add the phrase `Closes #ISSUENUMBER` to the description above

*If you are having difficulty fixing build errors, first consult <https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions> carefully, especially ["Common Build Errors"](https://github.com/programminghistorian/jekyll/wiki/Making-Technical-Contributions#common-build-errors). Then contact the technical team if you need further help.*


anisa-hawes:
Hello @hawc2. This PR is ready for you to review.  Apologies for my slow progress. I believe it now resolves the Issues identified in #2707. 

I've taken the following actions:

- [x] Removed `title =` and `subtitle =` where hard-coded within `labs()` argument
- [x] Revised l.179 to explain how titles and subtitles can be hard-coded alongside other descriptive text
- [x] Added figure numbers into captions (more consistent with other lessons)
- [x] Corrected the total number of Tweets at lines 205, 253, 304, 168 and 171
- [x] Replaced all three figures so that they do not feature titles, and the correct numbers are shown in their labels

hawc2 (approved):
This was a tough one to follow looking back at the issue ticket, but it seems like you fixed everything here @anisa-hawes!

PR #3035, opened 2023-09-29
Publish analyse donnees tabulaires r

marie-flesch:
I am publishing "Analyser des données tabulaires avec R". I've uploaded the 2 images and the lesson files, and added the French translation of the author bio to the authors yml.


anisa-hawes:
Hello @marie-flesch. I made some small adjustments to the numbering at lines 168 and 180 to solve a problem that sometimes occurs where elements (code, links or images) interrupt a numbered sequence. The sequence was reading 1., 1., ... instead of 1., 2., ....

Of course, we want the sequence to continue rather than repeat. The solution is usually to add a backslash after the number like this: `1\. 2\.`  But what I noticed when I implemented this https://github.com/programminghistorian/jekyll/pull/3035/commits/3cbf02db5b8701b969cf8be1a33e133761210d78 was that the numbers weren't indented. I tried removing the `\` and adjusting the line breaks instead. 

Either option gives us the number sequence we want but doesn't allow us indented numbers. You will notice that [the numbers in the Solutions subsection](https://deploy-preview-3035--ph-preview.netlify.app/fr/lecons/analyse-donnees-tabulaires-r#solutions) are not indented as in [the Mise en Pratique subsection directly above](https://deploy-preview-3035--ph-preview.netlify.app/fr/lecons/analyse-donnees-tabulaires-r#mise-en-pratique). What do you think?

Otherwise, I think it is great ✨

marie-flesch:
Thank you for the adjustments! I think the numbering is fine honestly (it's the same way in the original lesson anyway)

anisa-hawes:
Agreed. Looks like their work-around was to use A., B., instead of 1., 2., ...

anisa-hawes (approved):

marie-flesch:
thank you Anisa!

Issue #3353, opened 2024-09-07
Improve ergonomics of serial lessons -- add link to previous/next lesson at the bottom of the page


matgille:
Hi everyone, 

## If you are reporting a technical problem in the Programming Historian site or in the repository, please try to provide us with the following information:

The aim of this issue is to improve the ergonomics and consultability of the lessons that come in series. A banner at the top of the page is used to navigate between lessons in the same series, but it would be useful to add one at the bottom of the page, to enable the reader to move easily to the next lesson. 

I've searched the repository's discussion history and can't find anything about this. 

The simplest solution is to duplicate lines [229](https://github.com/programminghistorian/jekyll/blob/gh-pages/_layouts/lesson.html#L229)  to 239 of the [lesson.html](https://github.com/programminghistorian/jekyll/blob/gh-pages/_layouts/lesson.html) template, at the bottom of the page:

```html
  {% if page.sequence %}
    <div class="series-warning alert alert-info">
    {{ site.data.snippets.series-note[page.lang] }} {{ page.series_total }} - {{ site.data.snippets.lesson-sequence[page.lang] }} {{ page.sequence }}  
      {% if page.previous %}
        | <a href="{{page.previous}}"> {{ site.data.snippets.previous-lesson[page.lang] }}</a>
      {% endif %}
      {% if page.next %}
        | <a href="{{page.next}}"> {{ site.data.snippets.next-lesson[page.lang] }}</a>
      {% endif %}
    </div>
  {% endif %}
```

I'm still unsure about the exact location of the element: above the authors' presentation or below? Or below the citation division ? 

What do you think ?

All the best, 

Matthias

matgille:
Preview example: 
https://deploy-preview-3354--ph-preview.netlify.app/en/lessons/working-with-web-pages

anisa-hawes:
Thank you for this suggestion, @matgille. 

We are about to launch a project to renew our technical infrastructure and improve our website user interface, so I think this isn't the right moment to implement this change.

However, I will add this to our list of suggestions for enhancements to user experience. In the coming weeks, we will be inviting suggestions and ideas from the Project Team, so you will also have the opportunity to contribute this idea (and any others you have!) at that time.

I'm closing this Issue for now, but I am grateful for your contribution. Thanks again ☺️

PR #3354, opened 2024-09-07
Add link to previous/next lesson after citation

matgille:
Possible solution to  #3353. 

First possibility: to add the next/previous lesson division after the suggested citation. 

Maybe some kind of rule would improve readability ?

### DIFF
--- a/_layouts/lesson.html
+++ b/_layouts/lesson.html
@@ -288,6 +288,9 @@ All lesson metadata and alerts should follow the convention of pulling appropria
   <div class="content">
     {{ content }}
   </div>
+  
+  
+
 
   {% include author-info.html %}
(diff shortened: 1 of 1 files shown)


anisa-hawes:
Thank you for this suggestion, @matgille. 

We are about to launch a project to renew our technical infrastructure and improve our website user interface, so I think this isn't the right moment to implement this change.

However, I will add this to our list of suggestions for enhancements to user experience. In the coming weeks, we will be inviting suggestions and ideas from the Project Team, so you will also have the opportunity to contribute this idea (and any others you have!) at that time.

I'm closing this PR for now, but I am grateful for your contribution. Thanks again ☺️

